Issue #34, opened 2019-10-07
Extend dataset to cover substantial time range for multiple instruments


Lercerss:
As a user, I want to visualize data from more than one instrument and many days.

According to our current approximation, for 253 trading days (a full year) with snapshots every 10 seconds, the dataset size for SPY is around 40 GB.
We can enrich this by adding other instruments with smaller time ranges easily, for approximately less than 5 GB.

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 3 |
| Priority     | High |
| Risk         | High |

# Tasks
- Extend current SPY data to a full day
- Replicate full day's data over 253 trading days
- Use data samples from AMZN, AAPL, MSFT, INTC and GOOG to create 2-3 weeks of data for these instruments.

# Demo steps
- Display list of instruments available
- Display data for SPY more than 6 months apart

## Customer signoff
- [ ] Pierre-Marc Fournier
- [ ] TA

[//]: # (Make sure to add the appropriate feature label to your issue)

chimano:
### Acceptance test
| Steps                    | Inputs | Expected Outputs                                          | Status |
|--------------------------|--------|----------------------------------------------------------|--------|
| Import the data using the extender | N/A    | N/A | N/A   |
| Open the database | N/A    | N/A |N/A   |
| Go to the messages collection | N/A    | There should be a day's worth of messages in it |PASS   |
| Go to the orderbooks collection | N/A    | There should be a day's worth of orderbooks in it |PASS  |

Issue #25, opened 2019-10-02
Parsing modify messages is ambiguous


Lercerss:
Parsing "modify" messages is not correctly handled right now or leads to ambiguous logic. We should clarify the algorithm required and compare our results with the pre-computed L2 orderbook file provided by LobsterData.

#5, #7

PR #56, opened 2019-10-22
[#25, #34] Extending sample to full days

Lercerss:
- Moved LobsterData-related code to lobster directory
- Added extender module to extend sample to full days
- Minor refactoring for running tests and lint on scripts

Importer now accepts an additional argument: `-e|--extend [DAYS, DUPLICATES]`
Both default to 1 when not provided

`DUPLICATES` controls how many times the sample is duplicated to create more orders on the book at any given time.

Example usage for 10 days, with messages duplicated once:
`python -m importer SPY_messages.csv "2012-06-21 00:00:00" SPY --extend 10 2`


core/scripts/lobster/extender.py
```diff
@@ -0,0 +1,173 @@
...
+            return m.share_quantity
+        elif m.message_type == MessageType.EXECUTE:
+            total += m.share_quantity
+        elif m.message_type == MessageType.DELETE:
+            return total + m.share_quantity
+        elif m.message_type == MessageType.MODIFY:
+            total += m.share_quantity
+
+    # Order has no NEW or DELETE, we will generate a delete for the same amount
+    return total * 2
```
chris-baur: could just set total to proper value and break for the two above cases, so there's a single return

core/scripts/lobster/extender.py
```diff
@@ -0,0 +1,173 @@
...
+        if m.message_type == MessageType.NEW_ORDER:
+            return m.share_quantity - total
+        elif m.message_type == MessageType.EXECUTE:
+            total += m.share_quantity
+        elif m.message_type == MessageType.DELETE:
+            return 0
+        elif m.message_type == MessageType.MODIFY:
+            total += m.share_quantity
+
+    return total
```
chris-baur: same idea here

core/scripts/lobster/extender.py
Lercerss: Actually here trying to use a single return would require an additional boolean flag to determine if a `NEW_ORDER` or `DELETE` was seen, as line 62 is only reached if `messages` only contains executes or modifies. This case requires doubling the summed up total.
I think adding breaks here would simply make the function more confusing and make maintenance more difficult.

Issue #40, opened 2019-10-09
Style the MVP display of OrderBook


alvyn279:
The current implementation of the orderbook as a display needs to be styled. The outside containers should all be organized correctly. 

#5

# Tasks
- [x] Make the main view look nicer by adding some type of navigation header with a single tab. Basically trying to replicate a dashboard view.
- [x] OrderBook: 
    - [x] #37
    - [x] Adjust colors to make it nicer

PR #58, opened 2019-10-22
#40 +. Using FE template for graphelier tool. UI tweaks, and small code changes.

chris-baur:


docker-compose.yml
```diff
@@ -25,4 +25,8 @@ services:
     ports:
       - "27017:27017"
     volumes:
-      - "./core/db/data:/data/db"
+      - type: "volume"
+        source: "mongodata"
+        target: "/data/db"
+volumes:
```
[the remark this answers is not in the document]
chris-baur: yes, Lucas was gonna talk with chirac to see what will be done for everyones setup.

panoreak:
Waiting for the changes we talked about during the meeting. I will approve once those are implemented 👍

Lercerss:
It seems like most of the files you've modified have had their line-endings modified since they appear as a full diff, they are probably converted to CRLF instead of the unix LF. You should configure your git installation to automatically convert files to LF

app/src/services/OrderBookService.js
```diff
@@ -1,8 +1,9 @@
...
-    static getOrderBookPrices = (instrument, timestamp) => {
-        return httpClient.get(`/orderbook/${instrument}/${timestamp}`);
-    };
+import {httpClient} from './HttpClient';
+
+export default class OrderBookService {
+
+    static getOrderBookPrices = (instrument, timestamp) => {
+        return httpClient.get(`/orderbook/${instrument}/${timestamp}`);
+        //return httpClient.get(`/orderbook/SPY/1340285560000000000`);
```
alvyn279: please take this out

app/src/components/template/Dashboard.js
```diff
@@ -0,0 +1,94 @@
...
+import OrderBookSnapshot from '../OrderBookSnapshot';
+import { Styles } from '../../styles/Dashboard';
+
+/**
+ * Returns copyright component for footer
+ */
+const Copyright = () => {
+    const classes = Styles();
+    return (
+        <Typography variant="body2" color="textSecondary" align="center" className={classes.paddingCopyright}>
```
alvyn279: stuff like double quotes should be resolved by using the linting with our `eslintrc.json`. Your formatter is clearly not correct. I also encourage prop values to be set in curly braces, it's cleaner:

`variant={'body2'}`

app/src/components/template/Dashboard.js
```diff
@@ -0,0 +1,94 @@
...
+            </Link>{' '}
+            {new Date().getFullYear()}
+            {'.'}
+        </Typography>
+    );
+};
+
+/**
+ * Returns main componenet for the template
+ */
```
alvyn279: jsdocs should preferably be of the format: 
```
/**
* @description functionality
* @param foo first param
* @param bar second pram
*/
const functionName = (foo, bar) => {
   // ...
};
```

app/src/styles/Dashboard.js
```diff
@@ -0,0 +1,92 @@
+import { Colors } from './App';
+import { makeStyles } from '@material-ui/core/styles';
+
+const drawerWidth = 240;
```
alvyn279: `const DRAWER_WIDTH = 240`

chris-baur:
> @Lercerss 
> 
> It seems like most of the files you've modified have had their line-endings modified since they appear as a full diff, they are probably converted to CRLF instead of the unix LF. You should configure your git installation to automatically convert files to LF

So, i found this thread. Is this what I need to do? and to what degree?
I get that i need to change from crlf to lf, but perform this on all the files ? on my setup for future changes only? https://stackoverflow.com/questions/2517190/how-do-i-force-git-to-use-lf-instead-of-crlf-under-windows

Lercerss:
> > @Lercerss
> > It seems like most of the files you've modified have had their line-endings modified since they appear as a full diff, they are probably converted to CRLF instead of the unix LF. You should configure your git installation to automatically convert files to LF
> 
> So, i found this thread. Is this what I need to do? and to what degree?
> I get that i need to change from crlf to lf, but perform this on all the files ? on my setup for future changes only? https://stackoverflow.com/questions/2517190/how-do-i-force-git-to-use-lf-instead-of-crlf-under-windows

You would rather want to force LF over CRLF so I don't think that thread is what you want. Also I think there's a proper setting you can modify instead of using gitattributes, I believe your fix would be to set `git config core.autocrlf=input`

app/src/components/template/Dashboard.js
chris-baur: material UI demos uses like this. I didnt know that it wasnt correct. I have eslint extension installed, still isnt showing as a problem.

.gitignore
```diff
@@ -0,0 +1,5 @@
+#windows docker compose file
+docker-compose_windows.yml
```
alvyn279: 👍

docker-compose.yml
```diff
@@ -1,28 +1,28 @@
...
-      - "mongo"
-  mongo:
-    build:
-      context: ./core/db
-    container_name: "graphelier-db"
-    ports:
-      - "27017:27017"
-    volumes:
-      - "./core/db/data:/data/db"
+version: "3.4"
```
Lercerss: Line-endings issue in this file?

chris-baur: I guess there was an issue for this one as well. Is there a way for me to remove it from being "modified"

Lercerss: You can try doing a `git checkout origin/master -- docker-compose.yml` or something similar

Issue #59, opened 2019-10-23
Improve the speed of the scroller for choosing time


alvyn279:
Right now, when the horizontal scroller is moved, the `onChange` callback is called at every nanosecond change, and a `setState` along with it every time, which makes the action super slow. If we remove the `onChange` prop and only use `onChangeCommitted`, the performance is substantially better while conserving functionality.

```js
<Slider
        className={classes.timestampSlider}
        min={NANOSECONDS_IN_NINE_AND_A_HALF_HOURS}
        max={NANOSECONDS_IN_SIXTEEN_HOURS} 
        step={1}
        defaultValue={0}
        onChange={this.handleChangeTime} // remove this
        {...}
        onChangeCommitted={this.handleCommitTime} />
```

| Metric       | Value           |
|--------------|-----------------|
| Priority     | Medium |
| Risk         | Low |

# Tasks
- remove `onChange` prop for `Slider` and corresponding handlers

alvyn279:
solved in #60

PR #60, opened 2019-10-23
[#59] removed inefficient slider callbacks/setStates

alvyn279:
#59


alvyn279:
end result: blocked rendering of `TimestampOrderBookScroller` in cases where bids and asks did not change. Slider now moves seamlessly with a lot of data in the order book, while updating the nanosecond string label. Ready for review

Issue #55, opened 2019-10-22
Cache dependencies in CI build, keep going when tests/lint steps fail


Lercerss:
As a developer, I want to have quicker build and see all test/lint results even on a failed build. Quicker builds allow for faster response to errors while seeing all testing or linting results at once help address all issues found at once.

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 2 |
| Priority     | Medium |
| Risk         | Low |

# Tasks
- Cache container dependencies to make commands such as `go get` and `npm install` execute faster.
- Separate test & lint build steps by component (app, service, scripts) into jobs and let other jobs run even if one component's job has failed.

# Demo steps
- Builds complete in <4 mins
- Test failures on the app do not abort the build early

## Customer signoff
- [ ] Pierre-Marc Fournier

[//]: # (Make sure to add the appropriate feature label to your issue)

chimano:
### Acceptance test
| Steps                    | Inputs | Expected Outputs                                          | Status |
|--------------------------|--------|----------------------------------------------------------|--------|
| Go to CI host | N/A    | N/A | N/A   |
| Go to the latest build triggered by a PR | N/A    | The duration of the build should be less than 4 minutes. The build should be split into multiple jobs |PASS   |

PR #61, opened 2019-10-23
[#55] Added golangci config

Lercerss:
- Modified circleci config to run jobs by component
- Moved README contents
- Added test step for service

Linting for the service is now executed by GolangCI instead of a step in CircleCI. Code styling is still enforced according to `go fmt` standards. I will add it to required checks on PRs.

Failed builds for a component should no longer prevent the build step for other components from executing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2
 jobs:
-    build:
+    build_app:
         machine:
             image: ubuntu-1604:201903-01
         steps:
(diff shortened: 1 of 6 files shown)


chimano:
Should we allow other components to be built and tested if a component's tests fail? It'll take up more of our build minutes for a PR that will not get merged.

panoreak:
> Should we allow other components to be built and tested if a component's tests fail? It'll take up more of our build minutes for a PR that will not get merged.

However, if the PR has multiple issues and can make multiple tests fail, it would be more useful to know ALL of the issues in one build, and fix them all in one commit. Otherwise, if we can only see one failed test at a time, it may take multiple builds to get through all of the failed test cases.

Issue #29, opened 2019-10-02
Display order book price-level quantities boxes with a relative size


alvyn279:
We should observe order book quantities being displayed as relatively-sized boxes on the horizontal axis.
#27 

# Tasks
- [x] Figure out a mathematical approach to determine the relative size based on fixed witdh

alvyn279:
merged in #64

PR #64, opened 2019-10-23
[#29] Relative box sizes for quantities

riyaGit:
This following is a sample of the implementation:
- For small quantities, a tooltip is used to show the quantity  
Screen Shot 2019-10-23 at 4 49 53 PM

### DIFF
--- a/app/src/components/Order.js
+++ b/app/src/components/Order.js
@@ -7,10 +7,10 @@ import {Styles} from '../styles/Order';
 class Order extends Component {
 
     render() {
-        const {classes, type, quantity} = this.props;
-
+        const {classes, type, quantity, maxQuantitySum} = this.props;
+        const quantityBoxSize = (quantity / maxQuantitySum) * 100;
         return (
-            <span className={classNames(classes.rectangle, type === 'bid' ? classes.bid : classes.ask)}>
+            <span className={classNames(classes.rectangle, type === 'bid' ? classes.bid : classes.ask)} style={{width:`${quantityBoxSize}%`}}>
                 {quantity}
             </span>
         );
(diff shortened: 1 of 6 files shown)


app/src/components/TimestampOrderBookScroller.js
```diff
@@ -59,7 +60,19 @@ class TimestampOrderBookScroller extends Component {
             });
         });
 
-        this.setState({listItems}, () => {
+        if (listItems.length > 0){
+            listItems.map(listItem =>{
```
Lercerss: We can probably do this in a single pass in the `map`s on lines 55 & 47

riyaGit: sure

app/src/components/TimestampOrderBookScroller.js
```diff
@@ -83,7 +96,8 @@ class TimestampOrderBookScroller extends Component {
     };
 
     render() {
-        const {listItems} = this.state;
+        const {listItems, maxQuantitySum} = this.state;
+        const leeway = maxQuantitySum + maxQuantitySum*(35/100);
```
alvyn279: please rename this, and extract 35 % as a constant to express leeway

app/src/components/Order.js
```diff
@@ -3,16 +3,24 @@ import {withStyles} from '@material-ui/core/styles';
...
         return (
-            <span className={classNames(classes.rectangle, type === 'bid' ? classes.bid : classes.ask)}>
-                {quantity}
-            </span>
+            <Box className={classNames(classes.rectangle, type === 'bid' ? classes.bid : classes.ask)} style={{minWidth:`${quantityBoxSize}%`, maxWidth: `${quantityBoxSize}%`}}>
+                {quantityBoxSize > 1.5 ?
+                    (
+                        <Typography className={classes.text}>{quantity}</Typography>
+                    ) : (
+                        <Tooltip title={quantity} placement='cursor'><span style={{width: '100%', height: '100%'}}/></Tooltip>
```
alvyn279: can you put the width and height the stylesheet ?, would probably need to use `className`

Issue #53, opened 2019-10-20
Refactor fetch and storage for Orderbook


alvyn279:
We now want the current Orderbook state to be stored in the Redux store. This way, when a price level data changes, there is no need to re-render the entirety of price level components.

# Tasks
- Move away from data passing in `props`
- PriceLevel components should listen to changes in the store, that contains all the most up-to-date bids and asks for a price level.
- Highest level component should keep track of all the price levels to then map them to PriceLevel components

Highest level component will decide:
- if the response makes it so that the max quantities changes, re-render the whole orderbook
- if the response does not change the max quantity, spread changes to redux store, and only update a single price level component.

This will significantly improve the performance of the app, and will only re-render components that need to be re-rendered, in the context of moving message per message (#52)

[UPDATE]: To obtain better performance, one can just block the re-rendering at `PriceLevel` component with a `shouldComponentUpdate()`. This will avoid using Redux at all.

alvyn279:
```
====== REFACTOR - FLOW FOR MESSAGE INCREMENTS =======


PriceLevelComponent: price as native props, orderBookEntry & maxQuantity as redux props/state
orderBookEntry: {
	price: 103.4,
	orders: [...],
	type: 'ask',
	isMiddle: 'false',
}

CASE new date/timestamp chosen => requests completely new orderbook (setState, re-render of the whole table)

CASE move messages by increments/decrements of X {

	received at least one non-existing price level: {
		append to orderBook listItems in store / modify orderBook list items price levels
		O(n) scan to update max quantity
		detect change in store listItems length in OrderBookSnapshot an setState from there
	}
	

	received EXISTING price level(s) ONLY: {
		update price level(s) in store listItems in store
		update maximum quantity (by doing an O(n) pass on orderBook listItems in the store)
		perform a re-render ONLY on the existing PriceLevelComponent
	}
}
```

alvyn279:
From the discussion during the meeting, only we need `OrderBookSnapshot` component to be connected to Redux store and passing the information through props.

This allows for better maintainability and code clarity for `PriceLevel` component.

alvyn279:
Solved in #84

PR #68, opened 2019-10-26
[#53] Refactored orderbook storage to redux store

alvyn279:
Addressed: 
- Moved away from props passing to a global store for `orderBook`

```js
// Redux STORE:
orderBook: {
    listItems: {
        "103.4": {
	    price: 103.4,
	    orders: [...],
	    type: 'ask',
	    isMiddle: 'false',
        }, 
        "103.39": {...},
    },
    maxQuantity: 20900
}

```

- TODO: fix all tests
- TODO: function test with increment of a single message


app/src/components/OrderBookSnapshot.js
```diff
@@ -27,69 +30,67 @@ class OrderBookSnapshot extends Component {
...
+    handleChangeTime = (event, value) => {
+        let selectedTimeNano = parseInt(value);
+
+        this.setState({
+            selectedTimeNano: selectedTimeNano,
+            selectedTimeString: nanosecondsToString(selectedTimeNano),
+        });
+    };
+
+    /**
```
chris-baur: why is all this considered different ?

app/src/components/PriceLevel.js
```diff
@@ -6,26 +6,83 @@ import {Styles} from '../styles/PriceLevel';
 
 import Order from './Order';
 import {Box} from '@material-ui/core';
+import {connect} from 'react-redux';
+
+const MIN_PERCENTAGE_FACTOR_FOR_BOX_SPACE = 0.25;
```
chris-baur: we currently have a constants file, I believe it should be moved to there unless that file is only for constants we plan on reusing

app/src/components/OrderBookSnapshot.js
panoreak: I think this is reverting the line endings. I just opened my pr, and I have a huge diff like this too.

alvyn279:
After even further discussions, we will maintain a pass by props flow, except try to stop the rendering of the child components (PriceLevel) unless they are updated.

alvyn279:
Will open a new PR addressing more up-to-date changes needed

Issue #52, opened 2019-10-20
Add functionality for stepping forward and backwards at message level


alvyn279:
#28

When viewing the Orderbook, one should be able to step forward and backwards at message granularity.
Our request to the backend should take into consideration a timestamp along with an offset (in terms of messages).

When a message move is made:
- The orderbook should update up the price levels that have changed. It's **0 up to *offset_amount*** price levels that will need to be updated in the data structure used for the orderbook table. 

# Tasks
- [x] Add buttons to move forward and backwards at message-level
- [x] Send request for a particular timestamp and offset
- [x] Process response data and modify the values for listItems in OrderBookSnapshot

alvyn279:
TODO: integrate once #35 BE functionality is implemented.

alvyn279:
solved in #70

Issue #28, opened 2019-10-02
Add time dimension to order book display


Lercerss:
As a user, I want to select the time for which the order book is displayed. Furthermore, I want to be able to step randomly across time or step through time one message at a time.

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 5 |
| Priority     | High |
| Risk         | High |

# Tasks
- Allow moving through time in the front-end
- Serve deltas when requesting message steps from service
- Refresh display with new data

# Demo steps
- Open the order book display
- Select a different time in the future
- Select a different time in the past
- Notice that the order book display is refreshed with data for the selected time
- Move one message forward in time
- Move one message backward in time
- Notice that the order book is refreshed with data affected by the message

## Customer signoff
- [ ] Pierre-Marc Fournier

[//]: # (Make sure to add the appropriate feature label to your issue)

alvyn279:
Finished implementation in #70

chimano:
### Acceptance test
| Steps                    | Inputs | Expected Outputs                                          | Status |
|--------------------------|--------|----------------------------------------------------------|--------|
| Go to main page | N/A    | N/A | N/A   |
| Select a datetime to generate an orderbook | N/A    | A table should appear, representing the order book at the selected point in time.  | -   |
| Select another datetime to generate an orderbook | N/A    | A different table should appear.  | -   |
| Go forward in time by 1 message | N/A    | A row on the table should be modified.  | -   |
| Go backward in time by 1 message | N/A    | The modified row should get reverted.  | -   |

PR #70, opened 2019-10-26
Message stepping frontend

panoreak:
Added arrows for stepping forwards and backwards one message at a time.
Includes UI tests.

Task: [#52]
Story: [#28]


app/src/components/OrderBookSnapshot.js
```diff
@@ -1,42 +1,61 @@
...
      * @desc Handles the date change for the TextField date picker
      * @param event The event object that caused the call
      */
     handleChangeDate = (event) => {
-        let selectedDateNano = parseInt(dateStringToEpoch(event.target.value + ' 00:00:00'));
+        const { selectedDateNano, selectedDateString } = this.state;
+        let value = event.target.value;
+        let newDateNano;
+        if (value === ''){ // If user clears the date input
+            this.setState({
```
Lercerss: Why do you call `setState` here if the values have not changed? Won't that trigger a render?

app/src/components/OrderBookSnapshot.js
```diff
@@ -90,103 +109,161 @@ class OrderBookSnapshot extends Component {
...
+            .then(response => {
+                let { timeNanoseconds, dateNanoseconds } = splitNanosecondEpochTimestamp(response.data.timestamp);
+                this.setState({
+                    selectedDateNano: dateNanoseconds,
+                    selectedDateString: epochToDateString(dateNanoseconds),
+                    selectedTimeNano: timeNanoseconds,
+                    selectedTimeString: nanosecondsToString(timeNanoseconds),
+                    selectedDateTimeNano: response.data.timestamp,
+                    asks: response.data.asks,
+                    bids: response.data.bids,
```
Lercerss: This won't work with the backend implementation, as we are not sending a full orderbook back. Rather the asks and bids should be "upserted" into the current orderbook state

app/src/services/OrderBookService.js
```diff
@@ -1,8 +1,12 @@
...
\ No newline at end of file
+import {httpClient} from './HttpClient';
+
+export default class OrderBookService {
+
+    static getOrderBookPrices = (instrument, timestamp) => {
+        return httpClient.get(`/orderbook/${instrument}/${timestamp}`);
+    };
+
+    static getOrderBookPricesByMessageOffset = (instrument, timestamp, offset) => {
```
Lercerss: Maybe we should use price _levels_ here? Something like `getPriceLevelsByMessageOffset`

app/src/tests/component-tests/OrderBookSnapshot.test.js
```diff
@@ -77,3 +77,49 @@ describe('date and time picker functionality', () => {
         expect(selectedDateNano).toEqual(DATE_VALUE+TIME_VALUE);
     });
 });
+
+describe('navigating by message functionality', () => {
+    let mount, shallow;
+    const getOrderBookPricesSpy = jest.spyOn(OrderBookService, 'getOrderBookPrices').mockImplementation((instrument, timestamp) => Promise.resolve({ data: () => [] }));
+    const getOrderBookPricesByMessageOffsetSpy = jest.spyOn(OrderBookService, 'getOrderBookPricesByMessageOffset').mockImplementation((instrument, timestamp, offset) => Promise.resolve({ data: () => [] }));
```
Lercerss: If you want to start testing your implementation without the backend, you could have an orderbook structure here where you remove most of the price levels in the asks/bids response.

app/src/components/OrderBookSnapshot.js
panoreak: I wanted to disable clearing the date once it has been set. The TextField date picker now has the prop `value={selectedDateString}` so that I can set the date when we are going forwards/backwards by message and it rolls over to the next/previous day. When the user clicks on 'x' to clear the date on the date picker, it sets the value of the date picker to `''`. I could not find a way to disable the 'x' to clear the date, so my workaround was to set the state to its existing date, so that the value of the date picker would go back to what it was previously. I'm not sure if there's a better way to disable the clearing of the input.

The only solution I can think of is to allow users to clear the date.

app/src/components/OrderBookSnapshot.js
```diff
@@ -46,16 +50,33 @@ class OrderBookSnapshot extends Component {
      * @param event The event object that caused the call
      */
     handleChangeDate = event => {
-        const selectedDateNano = parseInt(dateStringToEpoch(`${event.target.value} 00:00:00`));
+        const { value } = event.target;
+        if (value === '') { // If user clears the date input
```
alvyn279: :+1:

app/src/components/TimestampOrderBookScroller.js
```diff
@@ -28,6 +39,12 @@ class TimestampOrderBookScroller extends Component {
         }
     }
 
+    onKeyUp = e => {
+        const { handleGoToPreviousMessage, handleGoToNextMessage } = this.props;
+        if (e.keyCode === LEFT_ARROW_KEY_CODE) handleGoToPreviousMessage();
+        else if (e.keyCode === RIGHT_ARROW_KEY_CODE) handleGoToNextMessage();
+    };
```
alvyn279: fancy

app/src/utils/date-utils.js
```diff
@@ -45,3 +49,31 @@ export const convertNanosecondsToUTC = nanosecondTimestamp => {
...
+
+/**
+ * @desc Given a nanosecond epoch timestamp, returns an object with date nanoseconds and time nanoseconds
+ * @param nanosecondTimestamp
+ * @returns {{timeNanoseconds: Number, dateNanoseconds: Number}} The timestamp split into its date nanoseconds and
+ * its time nanoseconds
+ */
+export const splitNanosecondEpochTimestamp = nanosecondTimestamp => {
+    // eslint-disable-next-line no-param-reassign
+    if (typeof nanosecondTimestamp === 'string') nanosecondTimestamp = parseInt(nanosecondTimestamp);
```
alvyn279: I think here you should actually fix it, (assign `nanosecondTimestamp` to a new variable and modify that variable)

app/src/utils/order-book-utils.js
```diff
@@ -94,3 +94,51 @@ export const ordersEquals = (ordersArray1, ordersArray2) => {
...
+        else if (listItems[price]) {
+            listItems[price].orders = orders; // Modify price level
+        } else {
+            listItems[price] = { // Add price level
+                price,
+                orders,
+                type: 'bid',
+            };
+        }
+    });
```
alvyn279: since this does the same thing, can it be extracted into a function **that looks like** the following:

```js

const updateListItem = (type, newArray, listItems) => {
   newArray.forEach(askOrBid => {
       const { price, orders } = askOrBid;
        if (orders.length === 0) delete listItems[price]; // Remove price level
        else if (listItems[price]) {
            listItems[price].orders = orders; // Modify price level
        } else {
            listItems[price] = { // Add price level
                price,
                orders,
                type,
            };
      }
```

app/src/components/TimestampOrderBookScroller.js
```diff
@@ -1,17 +1,28 @@
...
+import { LEFT_ARROW_KEY_CODE, RIGHT_ARROW_KEY_CODE, SNAPSHOT_INSTRUMENT } from '../constants/Constants';
+import OrderBookService from '../services/OrderBookService';
+
 const MIN_PERCENTAGE_FACTOR_FOR_BOX_SPACE = 0.35;
 
 class TimestampOrderBookScroller extends Component {
     middleReferenceItem = null;
 
+    componentDidMount() {
+        window.addEventListener('keyup', this.onKeyUp);
```
alvyn279: might want to remove this eventListener once this components unmounts. Or else it won't find the `onKeyUp` function if the component isn't displayed anymore and the user presses the right, left arrow keys

app/src/components/TimestampOrderBookScroller.js
```diff
@@ -28,6 +39,16 @@ class TimestampOrderBookScroller extends Component {
         }
     }
 
+    componentWillUnmount() {
+        window.removeEventListener('keyup', this.onKeyUp);
```
alvyn279: :+1:

Issue #76, opened 2019-10-30
Refactoring to include message start-of-day offset


Lercerss:
As was discovered while implementing the `/deltas` route part of #35, the backend needs to differentiate messages with equal timestamps. To accomplish this, a new unique field for messages should be introduced based on a combination of the message's current day and the number of messages that came before it for that day.

This will require changes to the `importer.py` to keep track and store the message's `sod_offset` value.

The service will need to return the `last_sod_offset` applied on the orderbook when returning the orderbook state. This value must be kept as part of app state on the frontend, then used for requests to deltas by message offset, and requests for list of messages.

The deltas routes should now accept: `/delta/{instrument}/{sod_offset}/{number_of_messages}`

The list of messages: `/messages/{instrument}/{sod_offset}/{number_of_messages}`

alvyn279:
solved in #82

PR #79, opened 2019-10-31
[#76] SOD offset now saved in DB

chimano:
- sod_offset added in messages collection
- last_sod_offset added in orderbooks collection

This is needed to add uniqueness to messages

### DIFF
--- a/core/scripts/models/order_book.py
+++ b/core/scripts/models/order_book.py
@@ -32,6 +32,7 @@ class OrderBook:
             MessageType.EXECUTE: self._do_execute,
             MessageType.IGNORE: lambda x: None
         }
+        self.last_sod_offset = 0
 
     def _do_new(self, msg: Message):
         o = Order(msg.id, msg.share_quantity, msg.price, msg.direction)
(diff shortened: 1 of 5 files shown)


core/scripts/models/order_book.py
```diff
@@ -93,9 +96,10 @@ def _do_execute(self, msg: Message):
         else:
             self._do_delete(msg)
 
-    def send(self, msg: Message):
+    def send(self, msg: Message, last_sod_offset: int = 0):
```
Lercerss: Here the sod_offset could simply be extracted from the message passed as parameter

chimano: oh right the drowsiness got to me

chimano: Oh wait no. I did it like this so that the tests would not be affected by it

Lercerss: How about adding the default value to the message model instead?

chimano: Smart man

Issue #65, opened 2019-10-23
Time interval displayed is in UTC


Lercerss:
App displays a time range from 9:30 AM to 4:00 PM which should be in EST, but it appears to be in UTC. The time range should correspond to the market open hours which are in EST for NASDAQ.

From browser logs, time `2012-06-21 14:28:47.394714904` is converted to `1340288927394715000` which according to [epochconverter](epochconverter.com) corresponds to `2012-06-21 14:28:47.394714904+0` while it should be converted to `2012-06-21 14:28:47.394714904-4`.

alvyn279:
Solved in #80

PR #80, opened 2019-10-31
[#65] Timestamp gets converted to UTC

chimano:
Added a rule in eslint to enforce the 1-prop-per-line rule.
Also, server was complaining about writing the header after writing the response so I changed the ordering on the backend.
Most important part of this PR is the `handleCommitTime` method in OrderBookSnapshot.js and the `convertNanosecondsToUTC` in date-utils.js.

### DIFF
--- a/app/src/components/OrderBookSnapshot.js
+++ b/app/src/components/OrderBookSnapshot.js
@@ -1,18 +1,32 @@
 import React, { Component } from 'react';
-import { withStyles, Typography, FormControl, TextField, Slider, Collapse, IconButton, Card } from '@material-ui/core';
+import {
+    withStyles,
+    Typography,
+    FormControl,
+    TextField,
+    Slider,
+    Collapse,
+    IconButton,
+    Card
+} from '@material-ui/core';
 import { Styles } from '../styles/OrderBookSnapshot';
-import { dateStringToEpoch, nanosecondsToString } from '../utils/date-utils';
+import {
+    dateStringToEpoch,
+    nanosecondsToString,
+    convertNanosecondsToUTC
+} from '../utils/date-utils';
 import TimestampOrderBookScroller from './TimestampOrderBookScroller';
 import classNames from 'classnames';
 import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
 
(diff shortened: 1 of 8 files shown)


Lercerss:
@chimano can you add a description?

chimano:
> @chimano can you add a description?

done

app/src/components/template/Dashboard.js
```diff
@@ -33,7 +33,11 @@ const Dashboard = () => {
     return (
         <div className={classes.root}>
             <CssBaseline />
-            <AppBar position={'absolute'} className={clsx(classes.appBar, open && classes.appBarShift)} theme={OBStyles.themeBackground}>
+            <AppBar 
+                position={'absolute'}
+                className={clsx(classes.appBar, open && classes.appBarShift)}
```
panoreak: Just to keep our code style consistent, you should use classNames:
```suggestion
                className={classNames(classes.appBar, open && classes.appBarShift)}
```

app/src/components/OrderBookSnapshot.js
```diff
@@ -31,12 +45,17 @@ class OrderBookSnapshot extends Component {
...
-        this.setState({
-            selectedDateNano,
-        }, () => { this.handleChangeDateTime(); });
+        this.setState(
+            {
+                selectedDateNano
+            },
+            () => {
+                this.handleChangeDateTime();
+            }
```
alvyn279: It's nice to have a linter on save but I don't know if the styling (in cases like above) is necessary. what do you think?

chimano: it's always better to have clear linting rules so that style remains consistent. Without defining it, it could lead to some editors formatting the same piece of code in different ways. The problem of large diffs can be solved with a more complete .eslintrc file

chimano: Perhaps if we use a preset, we won't have any more issues with larger diffs since there will be no ambiguity in terms of formatting.

alvyn279: Gonna look at this on my newest pr

PR #82, opened 2019-11-01
[#76] SOD offset now includes the dates timestamp

chimano:
- last_sod_offset now returned to the frontend when requesting the orderbook
- index created to optimize sorting of messages by sod_offset

### DIFF
--- a/core/db/db_setup.js
+++ b/core/db/db_setup.js
@@ -5,5 +5,7 @@ db.createCollection('orderbooks');
 db.createCollection('messages');
 
 db.orderbooks.createIndex({ instrument: 1, timestamp: 1 });
+
 db.messages.createIndex({ instrument: 1, timestamp: 1 });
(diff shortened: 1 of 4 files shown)


core/db/db_setup.js
```diff
@@ -5,5 +5,7 @@ db.createCollection('orderbooks');
 db.createCollection('messages');
 
 db.orderbooks.createIndex({ instrument: 1, timestamp: 1 });
+
 db.messages.createIndex({ instrument: 1, timestamp: 1 });
 db.messages.createIndex({ instrument: 1, sod_offset: 1 });
+db.messages.createIndex({ instrument: 1, sod_offset: -1 });
```
alvyn279: Why is this necessary?

chimano: We have to sort by sod_offset in both ways for pagination. Direction of the index matters for composite keys in mongodb. https://stackoverflow.com/questions/10329104/why-does-direction-of-index-matter-in-mongodb

Issue #51, opened 2019-10-16
Serve list of messages with pagination


Lercerss:
The service should be able to serve the list of messages based on an instrument and a time point, with pagination.

#50 

# Tasks
- Add route and request handler
- Read messages from database with a starting timestamp, offset and limit
- Add message serialization

PR #83, opened 2019-11-01
[#51] Serving messages with pagination

chimano:
- Backend now serves messages using sod_offset field
- Connector was changed to be an interface to allow for testing
- Handler now writes header before encoding json object because
it was giving warnings before

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
                   at: /go/src/graphelier
             - run:
                   name: Build service executable
-                  command: go get graphelier/core/graphelier-service
+                  command: go get -t graphelier/core/graphelier-service/...
             - save_cache:
                   key: v2-service-{{ .Branch }}
                   paths:
--- a/core/graphelier-service/api/hndlrs/msghandler.go
+++ b/core/graphelier-service/api/hndlrs/msghandler.go
@@ -1,0 +1,52 @@
+package hndlrs
+
+import (
+	"encoding/json"
+	"net/http"
+	"strconv"
+
+	"graphelier/core/graphelier-service/models"
+
+	"github.com/gorilla/mux"
+)
+
+// FetchMessages returns messages given an instrument and sod_offset with support for
+// pagination
+func FetchMessages(env *Env, w http.ResponseWriter, r *http.Request) error {
+	nMessages := r.URL.Query().Get("nMessages")
+
+	params := mux.Vars(r)
+	instrument := params["instrument"]
+	sodOffset := params["sodOffset"]
+
+	intNMessages, err := strconv.ParseInt(nMessages, 10, 8)
+	if err != nil {
+		intNMessages = 20
+	}
+
+	intSodOffset, err := strconv.ParseInt(sodOffset, 10, 64)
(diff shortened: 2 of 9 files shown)


core/graphelier-service/api/hndlrs/msghandler.go
```diff
@@ -0,0 +1,52 @@
...
+	"encoding/json"
+	"net/http"
+	"strconv"
+
+	"graphelier/core/graphelier-service/models"
+
+	"github.com/gorilla/mux"
+)
+
+// FetchMessages returns messages given an instrument and timestamp with support for
```
Lercerss: This comment seems outdated

chimano: Right

.circleci/config.yml
```diff
@@ -53,7 +53,7 @@ jobs:
                   at: /go/src/graphelier
             - run:
                   name: Build service executable
-                  command: go get graphelier/core/graphelier-service
+                  command: go get -t graphelier/core/graphelier-service/...
```
alvyn279: what does this line do ?

core/graphelier-service/api/hndlrs/msghandler.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	messagePage := models.MessagePage{PageInfo: paginator, Messages: messages}
+	if intNMessages < 0 {
+		reverseMessageOrdering(&messagePage)
+	}
+	w.WriteHeader(http.StatusOK)
+	err = json.NewEncoder(w).Encode(messagePage)
+	if err != nil {
+		return StatusError{500, err}
+	}
```
alvyn279: is it good practice to constantly re-affect by overriding the `err` var in this block? or creating multiple `err` variables with the same name, since the scope isn't quite defined

core/graphelier-service/db/mongo.go
```diff
@@ -88,3 +95,57 @@ func (c *Connector) GetMessages(instrument string, timestamp uint64, latestFullS
...
+	return results, nil
+}
+
+// Default Abs func for golang takes a float and returns a float. Need one for ints
+func abs(n int64) int64 {
+	if n < 0 {
+		return -n
+	}
+	return n
+}
```
alvyn279: Is ternary not a preferred option in go ?

.circleci/config.yml
Lercerss: This builds and installs every package under graphelier-service, along with its regular and test dependencies

core/graphelier-service/api/hndlrs/msghandler.go
Lercerss: I believe this is proper Go style for error handling

.circleci/config.yml
alvyn279: i guess my question is more for the `/...` at the end of the path

Lercerss: Recursively walks subdirectories

core/graphelier-service/db/mongo.go
chimano: There is no ternary operator in go

PR #84, opened 2019-11-02
Front-end bug fixes and ESLint setup

alvyn279:
- Fixed missing price in orderbook, crash on scroll, and two decimal price
- Refactored the handling and spreading of the raw orderbook data to make it less costly on re-renders
- Reformatted all the front-end source code to satisfy new custom coding style

For clarity regarding the bug fixes, only view changes to the first commit.


app/src/styles/PriceLevel.js
```diff
@@ -1,33 +1,35 @@
...
+        alignContent: 'flex-start',
+        padding: '5px 0px',
+        margin: '0px 10px',
+    },
+    price: {
+        float: 'left',
+        width: '50px',
+        textAlign: 'left',
+        paddingRight: 15,
+        color: Colors.darkGrey,
```
panoreak: nice 👍

app/.eslintrc.json
```diff
@@ -8,14 +8,36 @@
...
+        "array-callback-return": "off",
+        "import/prefer-default-export": "off",
+        "no-unused-vars": ["error", { "vars": "all", "args": "none"}],
+        "no-unused-expressions": "off",
+        "no-loop-func": "off",
+        "no-plusplus": "off",
+        "radix": ["error", "as-needed"],
+        "react/jsx-max-props-per-line": [1, { "maximum": 1 }],
+        "react/prop-types": "off",
+        "react/jsx-filename-extension": "off",
```
chimano: I like having the distinction between component files and other js files. Any reason why this was turned off?

app/.eslintrc.json
```diff
@@ -8,14 +8,36 @@
...
+        "react/jsx-indent" : ["error", 4],
+        "react/jsx-indent-props" : ["error", 4],
+        "react/jsx-curly-brace-presence": ["error", { "props": "always", "children": "ignore"}],
+        "react/jsx-props-no-spreading": "off",
```
chimano: Not a fan of this being off, can it be a warning instead. using the spread operator for props makes code harder to follow.

app/src/actions/actions.js
```diff
@@ -1,8 +1,6 @@
...
-    return {
-        type: ACTION_TYPES.SAVE_REACT_APP_NAME,
-        payload: name
-    };
-};
\ No newline at end of file
+export const saveReactAppName = name => ({
+    type: ACTION_TYPES.SAVE_REACT_APP_NAME,
+    payload: name,
+});
```
chimano: We don't need this anymore

alvyn279: @chimano actually I want to leave this to demonstrate how I want the connection with redux to be done for future store interactions.

chimano: Fair enough

app/.eslintrc.json
chimano: Nevermind then just turn it off

alvyn279: I agree with being strict with this rule, that's why i'm gonna just leave it as warning/error.

Issue #81, opened 2019-11-01
Extended dataset does not account for movements in price


Lercerss:
Fake messages added after the sample's end are placed at the same price levels as the beginning of the sample, which doesn't account for a shift in price over the sample's duration.

PR #90, opened 2019-11-06
[#81] Adjust price on extended samples

Lercerss:
- If message would lead to a price movement, drop all orders that would conflict with it first
- Prices are changed according to the top of book before the extended sample is started
- Added dependency on the `orderbook` file from Lobster, data can also be passed through the new `--top-of-book` argument to avoid downloading the file

### DIFF
--- a/core/scripts/models/order_book.py
+++ b/core/scripts/models/order_book.py
@@ -19,9 +19,9 @@ class Order:
 class OrderBook:
     def __init__(self, instrument):
         self.instrument = instrument
-        self.bid_book: Dict[float, List[Order]] = defaultdict(list)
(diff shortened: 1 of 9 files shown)


core/scripts/models/order_book.py
```diff
@@ -101,12 +101,34 @@ def send(self, msg: Message):
         self.last_time = msg.time
         self.last_sod_offset = msg.sod_offset
 
+    def conflicts(self, msg: Message):
```
chimano: Can you write a comment in this method kinda confused with the logic

Lercerss:
That's correct, data previously generated by the importer script was not valid

Issue #98, opened 2019-11-10
Service exits instead of returning invalid status code on error


Lercerss:
Service uses `log.Fatal` which calls exits the process instead of returning status codes such as 400, 500.

PR #99, opened 2019-11-10
[#98] Prevent exit on error in service

Lercerss:
`log.Fatal` leads to a call to `os.Exit(1)`, `Print` instead


core/graphelier-service/api/hndlrs/handlers.go
```diff
@@ -51,6 +51,6 @@ func (h CustomHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		default:
 			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 		}
-		log.Fatal(err)
+		log.Print(err)
```
pvacca97:
```suggestion
		log.Println(err)
```

pvacca97: i feel like this would just be cleaner in the terminal?

Issue #95, opened 2019-11-10
Convert uint64 to string before sending it to the frontend


chimano:
Javascript's Number can only be as high as 9,007,199,254,740,991 until it starts losing its precision. Since we are dealing with numbers bigger than that, they should be converted to a string before being sent to the frontend.

alvyn279:
Solved in #102

PR #102, opened 2019-11-10
[#95] Converted timestamp to int to keep precision

chimano:
- Importer previously lost precision when storing to the db because it
was storing floats for the timestamp field. Now it will store ints.

### DIFF
--- a/core/scripts/importer.py
+++ b/core/scripts/importer.py
@@ -27,7 +27,7 @@ EOD = 16 * 60 * 60 * 10**9  # 4:00 PM as ns
 
 
 def load(message_file, ob_file_path, start_time, instrument, extend, top_of_book):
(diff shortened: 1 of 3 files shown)


core/scripts/importer.py
```diff
@@ -37,8 +37,8 @@ def load(message_file, ob_file_path, start_time, instrument, extend, top_of_book
     for day in weekdays(start_timestamp, int(extend[0])):
         order_book = OrderBook(instrument)
         order_book.last_time = day
-        # day is a float, need to convert it to avoid floating point error
-        order_book.last_sod_offset = sod_offset_counter = int(day)
+
+        order_book.last_sod_offset = sod_offset_counter = day
```
pvacca97: Why do we remove int here?

pvacca97: Wait nvm I get it

Issue #54, opened 2019-10-20
Display list of messages


alvyn279:
Alongside the orderbook, we should have a table that lists all the messages related to the orderbook, with a highlight on the message that represents the current orderbook state.

#50 

#Tasks
- [x] Mockup
- [x] Make request for messages:
    - [x] Orderbook should be infinitely scrollable on top and on bottom (paging)
- [x] Highlight correct message with respect to current orderbook state (change highlight when moving message)

alvyn279:
Mockups will be needed for this.

Lercerss:
On the backend side we have the following information on messages: time in ns, message type, order id, price, quantity, direction
Message type is an integer enum defined under https://github.com/Lercerss/graphelier/blob/master/core/scripts/models/message.py#L6

Lercerss:
I think we should have the value for `direction` translated to buy/sell as the integer values are not very meaningful

riyaGit:
Agreed. The mockup has been updated to reflect that.

Lercerss:
> Agreed. The mockup has been updated to reflect that.

Thanks, looks good!

chimano:
Will the orderbook and messages be seen at the same time or will the user have to scroll away from the orderbook to view the list of messages? It looks like if we fit both within the screen it'll hide too many rows from the orderbook. Is this what we want?

PR #103, opened 2019-11-11
Message List #54

riyaGit:
Screen Shot 2019-11-10 at 7 25 35 PM


app/src/components/MessageList.js
```diff
@@ -0,0 +1,201 @@
...
+
+            const { timeNanoseconds, dateNanoseconds } = splitNanosecondEpochTimestamp(timestamp);
+            const date = epochToDateString(dateNanoseconds);
+            const time = nanosecondsToString(timeNanoseconds);
+
+            return (
+                <Box
+                    key={`${sod_offset} ${timestamp}`}
+                    className={lastSodOffset.toString() === sod_offset
+                        ? classes.currentMessageRow
```
panoreak: Nice! From what I understand, this highlights the last APPLIED message, right?

riyaGit: yes, the "correct message with respect to current orderbook state" -issue #54

Lercerss:
While function testing this branch, a backend bug for deltas going backwards. @chimano is going to push a fix for this

Issue #105, opened 2019-11-11
Scale application display size


Lercerss:
As a user, I want the application to be scalable to different screen sizes, right now it is hardcoded for ratios around 1080p.

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 1 |
| Priority     | High |
| Risk         | High |

# Tasks
- Modify the height distribution of the graph to be based on the view. Right now it's hardcoded. 

# Demo steps
- Should display the graph and the list of messages fine on 4k screens and 1080p.

riyaGit:
Resolved in PR #115

riyaGit:
### Acceptance test
| Steps                    | Inputs | Expected Outputs                                          | Status |
|--------------------------|--------|----------------------------------------------------------|--------|
| Go to main page while on a 1080p screen and Select a datetime to generate an orderbook  | N/A    | The orderbook  and the message list should appear proportional to the screen |PASS   |
| Go to main page while on a 4k screen and Select a datetime to generate an orderbook  | N/A    | The orderbook  and the message list should appear proportional to the screen |PASS   |

PR #115, opened 2019-11-14
Scale application display size #105

riyaGit:
1080p screen:
Screen Shot 2019-11-13 at 10 49 55 PM

4k Screen:
Screen Shot 2019-11-13 at 10 59 24 PM


riyaGit:
> I think instead of having pixel height we should look into simply using relative sizes from css:
> https://www.w3schools.com/cssref/css_units.asp
> namely `vh`.
> 
> That would eliminate the use of util functions returning pixel sizes.

Yes that actually works wonderfully!

Issue #113, opened 2019-11-13
List available instruments


Lercerss:
On the backend, add a new route to serve all distinct instruments for which data is available.
#77

PR #121, opened 2019-11-14
[#113] Instruments route

Lercerss:
Serve available instruments to let app know which values are valid

Route spec:
    path: `/instruments/`
    response format: `["SPY","AAPL","MSFT"]`, a json list of strings


Lercerss:
The ones with empty return are not being used in the given test suite, they are simply added to fulfill the db's interface

Issue #114, opened 2019-11-13
Add menu option for selecting an instrument


chris-baur:
#77

There must be a component for the user to be able to choose/select between different instruments.

# Tasks
- [x] Add a drop down select to choose from available instruments
- [x] Add service call for getting instruments
- [x] Modify services to include use of specified instruments

PR #122, opened 2019-11-18
[#114] Add menu option for selecting an instrument

chris-baur:


app/src/components/MessageList.js
```diff
@@ -75,13 +75,13 @@ class MessageList extends Component {
...
      */
     fetchInitialMessages = async () => {
-        const { lastSodOffset } = this.props;
+        const { lastSodOffset, instrument } = this.props;
         const nMessages = MESSAGE_LIST_DEFAULT_PAGE_SIZE * 2 + 1;
         const lowerSodOffset = lastSodOffset - BigInt(MESSAGE_LIST_DEFAULT_PAGE_SIZE - 1);
 
         try {
             const reverseMessagesResponse = await OrderBookService.getMessageList(
-                SNAPSHOT_INSTRUMENT,
```
Lercerss: You should remove the matching import, and the constant from the Constants file

.gitignore
```diff
@@ -2,4 +2,7 @@
 docker-compose_windows.yml
 
 #Visual studio folder
-**/.vs
+**/.vs
+
+#Webstorm folder
+.idea/*
```
Lercerss: This is more flexible
```suggestion
**/.idea
```

app/src/components/template/Dashboard.js
```diff
@@ -11,20 +11,32 @@ import {
...
 
         this.state = {
             open: false,
+            selectedInstrument: '',
+            instruments: [],
         };
+
+        const { instruments } = this.state;
+
+        OrderBookService.getInstrumentsList().then(response => {
```
Lercerss: Shouldn't this be in `componentDidMount`?

app/src/tests/component-tests/Dashboard.test.js
```diff
@@ -23,4 +23,12 @@ describe('Dashboard functionality', () => {
         wrapper.find(IconButton).first().simulate('click');
         expect(wrapper.state().open).toBe(false);
     });
+
+    it('renders a Dashboard component with drawer and simulate selecting instruments', () => {
```
Lercerss: Are you mocking the service call here? I don't see the change

app/src/components/MessageList.js
chris-baur: thanks for pointing out!

app/src/components/template/Dashboard.js
chris-baur: 👍

app/src/tests/component-tests/Dashboard.test.js
chris-baur: 👍

app/src/components/template/Dashboard.js
```diff
@@ -11,22 +11,37 @@ import {
...
+            selectedInstrument: '',
+            instruments: [],
         };
     }
 
+    componentDidMount() {
+        const { instruments } = this.state;
+        const newInstruments = instruments.slice();
+        OrderBookService.getInstrumentsList().then(response => {
+            response.map(value => {
```
alvyn279: In `axios` the response object from the server will hold the array at the following location in the response:
```
response.data;
```
Notice the `.data`. You would have to do `response.data.map(...)`

app/src/components/template/Dashboard.js
```diff
@@ -11,22 +11,37 @@ import {
...
+                newInstruments.push(value);
+            });
+        });
+        this.setState({ instruments: newInstruments });
```
alvyn279: asynchronousity is wrong here, the `setState` should happen in the `.then()` callback of your service call.

app/src/components/template/Dashboard.js
```diff
@@ -93,6 +116,31 @@ class Dashboard extends Component {
...
+                                                key={`menuitem-${value}`}
+                                                value={value}
+                                            >
+                                                {value}
+                                            </MenuItem>
+
+                                        );
+                                    })
+                                }
+                            </Select>
```
alvyn279: Should this instrument selector only be proper to the OrderBook snapshot component in our app.
i.e. only visible when selecting on the side bar:

image

Because right now it's visible at the root

chris-baur: hmm, i Put it app wide. outside of the container used currently for displaying our components based on routing. So currently you will see it no matter what page you are on. How would we prefer it?

alvyn279: think it makes more sense for it to be only visible once we click on orderbook tab

app/src/components/template/Dashboard.js
chris-baur: thanks for pointing out

app/src/components/template/Dashboard.js
chris-baur: 👍

app/src/components/template/Dashboard.js
chris-baur: kk 👍

chris-baur:
> 
> 
> Your branch includes some duplicate commits already found in master. Can you fix the commit history? It also makes the diff quite weird

I rebased off of master, as you or Alvyn suggested. I do not know how you could remove duplicate commits.

alvyn279:
@Lercerss @chris-baur wont a merge commit fix all of this

chris-baur:
@Lercerss ok, thank you.

Issue #128, opened 2019-11-20
Empty orderbook response is not correct


Lercerss:
Response for `orderbook` far in the future does not contain the keys `bids` and `asks`.

Error response is returned for `orderbook` far in the past:
```
mongo: no documents in result
```

Both of these responses break front-end expectations.

Issue #94, opened 2019-11-09
Remove dependency on snapshot interval in service code


Lercerss:
The service currently expects exactly 10**9 nanoseconds between orderbook snapshots in the database. We should remove this dependency on the snapshot interval to allow modifying this value for performance analysis.

This will allow us to vary the value in the `importer` script without modifying the backend code.

Lercerss:
We can store the snapshot interval in an additional db collection, on a per instrument basis

PR #129, opened 2019-11-26
[#94, #128] Store interval value in meta collection

Lercerss:
- Importer script maintains value per instrument
- Service reads and caches meta values on startup
	- This can also be refreshed by a GET to `/_refresh/`
- Default interval value in the script has been reduced to 0.1s from 10s
- Unified response format for `/orderbook/`, timestamps out of bounds return an empty orderbook

To allow migration of existing dataset, simply:
```
db.createCollection("meta")
db.meta.insertOne({instrument: "SPY", interval: NumberLong("10000000000")})
db.orderbook.insertOne({instrument: "SPY", bids: [], asks: [], timestamp: NumberLong("0"), last_sod_offset: NumberLong("0")})
```

### DIFF
--- a/core/graphelier-service/api/hndlrs/obhandler.go
+++ b/core/graphelier-service/api/hndlrs/obhandler.go
@@ -128,6 +128,5 @@ func RefreshCache(env *Env, w http.ResponseWriter, r *http.Request) error {
 		return StatusError{500, err}
 	}
 
-	w.WriteHeader(http.StatusOK)
 	return nil
 }
--- a/core/graphelier-service/api/hndlrs/obhandler.go
+++ b/core/graphelier-service/api/hndlrs/obhandler.go
@@ -2,6 +2,7 @@ package hndlrs
 
 import (
 	"encoding/json"
+	"log"
 	"net/http"
 	"strconv"
 
@@ -34,7 +35,6 @@ func FetchOrderbook(env *Env, w http.ResponseWriter, r *http.Request) error {
 	}
 	orderbook.ApplyMessagesToOrderbook(messages)
 
-	w.WriteHeader(http.StatusOK)
 	err = json.NewEncoder(w).Encode(orderbook)
 	if err != nil {
 		return StatusError{500, err}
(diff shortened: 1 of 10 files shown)


core/graphelier-service/api/hndlrs/obhandler.go
```diff
@@ -114,3 +114,17 @@ func FetchOrderbookDelta(env *Env, w http.ResponseWriter, r *http.Request) (err
...
+
+// RefreshCache : Updates the db connector's cache to reflect database state changes
+func RefreshCache(env *Env, w http.ResponseWriter, r *http.Request) error {
+	err := env.Connector.RefreshCache()
+	if err != nil {
+		return StatusError{500, err}
+	}
+
+	log.Println("Cache refreshed.")
+	json.NewEncoder(w).Encode(make(map[string]string))
```
golangcibot: Error return value of `(*encoding/json.Encoder).Encode` is not checked (from `errcheck`)

core/graphelier-service/api/routes.go
```diff
@@ -58,4 +58,10 @@ var routes = Routes{
 		"/instruments/",
 		hndlrs.CustomHandler{H: hndlrs.FetchInstruments},
 	},
+	Route{
+		"RefreshCache",
+		"GET",
+		"/_refresh/",
```
alvyn279: i get the use of the `_`, but this isn't really compliant with REST I believe. Can it just be a more descriptive and meaningful route name ?

Lercerss: This is purely a "debug" route, therefore I did not think RESTful concerns applied. I can make it more compliant by changing the verb to a POST but it would make it less user friendly as it would not be possible to call it directly from the browser

alvyn279: fair enough

core/graphelier-service/db/mongo.go
```diff
@@ -19,11 +19,32 @@ type Datastore interface {
 	GetMessagesWithPagination(instrument string, paginator *models.Paginator) ([]*models.Message, error)
 	GetSingleMessage(instrument string, sodOffset int64) (*models.Message, error)
 	GetInstruments() ([]string, error)
+	RefreshCache() error
 }
 
 // Connector : A struct that represents the database
 type Connector struct {
 	*mongo.Client
+	cache cache
```
chimano: When does this get populated initially?

Lercerss: [Right here](https://github.com/Lercerss/graphelier/pull/129/files#diff-1717a8460c15d66b9eab79dbc9668be6R64) in the function `NewConnection`

chimano: makes sense

PR #130, opened 2019-11-27
message click/#127

riyaGit:
Added a click listner on message list rows such that clicking on a message changes the order book state to reflect the message clicked on


.gitignore
```diff
@@ -6,3 +6,5 @@ docker-compose_windows.yml
 
 #Webstorm folder
 **/.idea
+
+core/scripts/env
```
Lercerss: We already have that in the `core/.gitignore` file

riyaGit: addressed!

Issue #104, opened 2019-11-11
Safeguard the use of BigInteger on application side


alvyn279:
We should use a library (https://www.npmjs.com/package/big-integer) to monitor the usage of big ints. Since JavaScript's `BigInteger` is not supported on all browsers, it is really important to safeguard its usage.

riyaGit:
Closed in #135

PR #135, opened 2019-12-02
Safeguard BigInt in JS #104

riyaGit:
Added big-integer library (https://www.npmjs.com/package/big-integer) to safeguard BigInteger type in js.


riyaGit:
> Why are the `big-integer` source code files pushed ?

my bad! Fixed!

Issue #107, opened 2019-11-13
Create production environment for react app


chimano:
For User Story #62 

Set up an Nginx image that serves the react app for our production environment.

- Modify the current dockerfile for the frontend or create a new one to have an nginx instance serving the frontend application
- Write the Nginx configuration

chimano:
Implemented in #136

PR #136, opened 2019-12-06
Set up prod for frontend #107

chimano:
Set up a basic nginx server to serve the react application

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -5,13 +5,14 @@ COPY package*.json /app/
 FROM base AS builder
 RUN npm install
 COPY . .
+RUN npm run build
 
 FROM builder AS dev
-CMD [ "npm", "start" ] 
+CMD [ "npm", "start" ]
 
-FROM builder AS test
-ENV CI true
-CMD [ "npm", "test" ]
-
-FROM builder AS lint
-CMD [ "npm", "run", "lint", "src/" ]
+FROM nginx:alpine AS prod
+COPY --from=builder /app/build /usr/share/nginx/html
+RUN rm /etc/nginx/conf.d/default.conf
+COPY nginx/nginx.conf /etc/nginx/conf.d
+EXPOSE 80
+CMD ["nginx", "-g", "daemon off;"]
--- a/app/nginx/nginx.conf
+++ b/app/nginx/nginx.conf
@@ -1,0 +1,9 @@
+
+server {
+    listen       80;
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html;                 
+    }
+}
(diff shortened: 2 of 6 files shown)


app/Dockerfile
```diff
@@ -5,13 +5,14 @@ COPY package*.json /app/
 FROM base AS builder
 RUN npm install
 COPY . .
+RUN npm run build
```
Lercerss: This will result in additional build time for the dev image, do we want that?

app/Dockerfile
```diff
@@ -5,13 +5,14 @@ COPY package*.json /app/
...
+CMD [ "npm", "start" ]
 
-FROM builder AS test
-ENV CI true
-CMD [ "npm", "test" ]
-
-FROM builder AS lint
-CMD [ "npm", "run", "lint", "src/" ]
\ No newline at end of file
+FROM nginx:alpine AS prod
```
Lercerss: Could you modify the readme to indicate how to run this image?

app/Dockerfile
chimano: Frontend image only gets built once for devs. If I run that command as prod I would also need to run npm install to get all dependencies. By building it at the build step I can keep the prod image very small by skipping the installation of dependencies. And since the frontend image does not get built that often I find that the added amount of time it takes to build the frontend image is well worth the space we'll be saving in the prod image size

app/Dockerfile
chimano: Sure

app/nginx/nginx.conf
```diff
@@ -0,0 +1,9 @@
+
+server {
+    listen       80;
+    location / {
+        root   /usr/share/nginx/html;
+        index  index.html index.htm;
+        try_files $uri $uri/ /index.html;                 
+    }
+}
```
alvyn279: so this would serve the react app on port 80?

docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
   graphelier-app:
     build:
       context: ./app
-      target: dev
+      target: prod
```
alvyn279: So from my understanding we'll need to remove `docker-compose.override.yml` from the prod env to activate prod ?

app/nginx/nginx.conf
chimano: yep

docker-compose.yml
chimano: no so if you just docker-compose without passing in a file it'll combine both docker-compose.yml and docker-compose.override.yml files. However, if we specify the file, its contents will not get merged with the override file. This allows us to run a dev environment with `docker-compose build && docker-compose up` and a prod environment with `docker-compose -f docker-compose.yml build && docker-compose -f docker-compose.yml up`

alvyn279: ok

Issue #134, opened 2019-11-28
Address Iteration 5 comments


Lercerss:
Based on feedback from last Iteration, we should improve a few things according to the grading scheme. 

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 3 |
| Priority     | High |
| Risk         | Low |

# Tasks
- [X] Stories should link to their respective tasks and vice versa. We currently only have traceability from tasks to stories
- [x] Add class diagram for the scripts
- [x] Improve existing architecture diagram and make it more rigorous
- [x] Place all mockups in the wiki
- [x] Increase logging to be more verbose and use different log levels
- [x] Add comments to service/scripts tests, alternatively use BDD testing for all components
- [x] Review variable naming to be more descriptive
- [x] Test coverage and linting reports on builds

[//]: # (Make sure to add the appropriate feature label to your issue)

Lercerss:
![scripts_class_diagram](https://user-images.githubusercontent.com/34099471/71497826-ea6bf280-2827-11ea-86de-303880e89aeb.jpg)
Class diagram representing the current state of the scripts

Lercerss:
![data_view](https://user-images.githubusercontent.com/34099471/71857326-1ce1df00-30b5-11ea-8f99-646fe3786df1.jpg)
Data view representing the structure of the MongoDB collections

PR #138, opened 2019-12-29
[#134] BDD tests using mamba for scripts

Lercerss:
- Refactored tests into spec file to be self-documenting
- Added mamba as a test runner, with verbose output in CI

To run tests using mamba:
```
source venv/bin/activate
mamba `find . -name *.spec.py` --format=documentation
```

I have also refactored the scripts to use a logger, which writes to file and outputs to stdout.
Single letter variables caused us to lose marks in Release 1 therefore I renamed most of them.


Lercerss:
I found this lightweight BDD test runner called `mamba` for python, let me know what you guys think of the tests format. I found it was quite intuitive to use along with the `sure` assertion library.
Here is an example output using the `--format=documentation` option:
```
OrderBook:
  ✓ it receives a new order message
  when there exists an order on the book
    ✓ it deletes an existing order
    ✓ it modifies an existing order
    ✓ it executes a trade on an existing order
    ✓ it resets priority on orders upon receiving a modify with negative qty
    ✓ it detects possible conflicts for a given message
    ✓ it confirms the validity of a message

7 examples ran in 0.0454 seconds
```
I am going to leave this PR as draft until I convert all the existing test suites to spec files

Lercerss:
Test suites:
```
LobsterMessageParser:
  ✓ it parses sample messages
  ✓ it offsets messages from the start of given day

TopOfBook:
  ✓ it returns the proper side based on the given direction
  ✓ it can be subtracted

Extender:
  ✓ it yields weekdays starting from a given date
  ✓ it mixes Placement objects into a list
  when calculating properties of orders based on given messages
    ✓ it finds the initial share quantity from a NEW_ORDER
    ✓ it finds the initial share quantity without a NEW_ORDER
    ✓ it finds if an order is unfilled at the end of the sample
    ✓ it adjusts the quantities for orders with no NEWs or DELETEs
  when using the parser
    ✓ it parses the csv into Messages
    ✓ it mixes fake messages into the sample
    ✓ it yields a copy of the original sample first
    to make a cyclic sample
      ✓ it adds NEW_ORDER messages for orders missing one
      ✓ it adds DELETE messages for unfilled orders at the end of sample

OrderBook:
  ✓ it receives a new order message
  ✓ it confirms the validity of the sample fixture
  when there exists an order on the book
    ✓ it deletes an existing order
    ✓ it modifies an existing order
    ✓ it executes a trade on an existing order
    ✓ it resets priority on orders upon receiving a modify with negative qty
    ✓ it detects possible conflicts for a given message
    ✓ it confirms the validity of a message

23 examples ran in 0.0412 seconds
```

Issue #144, opened 2020-01-12
Docker exposes 2 ports for the frontend container when it shouldn't


chimano:
When running the dev environment, the production port also gets exposed when it shouldn't. This is most likely due to the behavior of overriding the 'ports' variable in the docker-compose yaml files.

It exposes ports 3000 and 80 for the frontend container when it should only expose port 3000

chimano:
Fixed in #145

PR #145, opened 2020-01-12
[#144] Fixed multiple ports being exposed

chimano:
- Running the dev environment now only exposes port 3000


chimano:
> actually before merging could you include the public facing ip in the README

Done

PR #152, opened 2020-01-21
[#134] Verbose logging with log levels for service

Lercerss:
Using https://github.com/sirupsen/logrus as a logger due to its ease of use and similar API to built-in `log` module.

Added logs to different handlers based on behavior. Also added logs to mongo calls to record performance.

Log format looks like this:
```
[TRACE]: 2020-01-20T23:58:49Z - mongo/RefreshCache took 497.982µs
```

### DIFF
--- a/core/graphelier-service/utils/utils.go
+++ b/core/graphelier-service/utils/utils.go
@@ -1,5 +1,11 @@
 package utils
 
+import (
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
 // Abs : Default Abs func for golang takes a float and returns a float. Need one for ints
 func Abs(n int64) int64 {
 	if n < 0 {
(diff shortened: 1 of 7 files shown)


Lercerss:
The timer could be overhauled to publish metrics pretty easily I believe

core/graphelier-service/utils/utils.go
```diff
@@ -1,9 +1,31 @@
...
 }
+
+func DebugTimer(call interface{}) func() {
+	start := time.Now()
+	return func() {
+		log.Debugf("%s took %v\n", call, time.Since(start))
+	}
+}
+
+func TraceTimer(call interface{}) func() {
```
chimano: How come you made `call` a generic interface rather than a string?

Lercerss: The idea was to make it more generic where a struct could be passed for printing, though I realize the format string uses `%s` and therefore can only accept strings. I think I will change the `interface` parameter to string and offload the formatting work to the caller of Timer.

Issue #141, opened 2020-01-09
Serve order information based on order id


Lercerss:
Preprocess order data, investigate storage/processing solutions and serve as a new route based on a given order id and _as-of_ timestamp.

# Tasks
- [x] Investigate algorithm and storage solutions for order information
- [x] Add new route which reads and serves order information based on order id and _as-of_ time

Story #110

chimano:
OrderInfo Architecture
![OrderInfo_Arch](https://user-images.githubusercontent.com/26588048/73128058-c40cb480-3f97-11ea-8484-aea44c0f5aa2.png)

chimano:
Handler tests: [link](https://github.com/Lercerss/graphelier/blob/master/core/graphelier-service/api/hndlrs/orderhandler_test.go)
Model tests: [link](https://github.com/Lercerss/graphelier/blob/master/core/graphelier-service/models/orderinfo_test.go)

PR #150, opened 2020-01-18
[#134] Utils, comments and refactoring for go test

Lercerss:
- Added auto-generated mock for the mongo connector using mockgen
- Refactored common test code into test_utils
- Added comments to detail steps in tests and expected behavior

PR #154, opened 2020-01-21
[#141] OrderInfo route added

chimano:
- Backend now serves the state of an order at any given time
- New index added to Mongo


core/graphelier-service/api/hndlrs/orderhandler.go
```diff
@@ -0,0 +1,66 @@
...
+		return StatusError{404, errors.New("Order was not found")}
+	}
+	orderInfoBuilder := &models.OrderInfoBuilder{}
+	orderInfoBuilder.
+		WithID(intOrderID).
+		WithInstrument(instrument).
+		WithMessages(&messages).
+		WithTimestamp(intTimestamp)
+	orderInfo := orderInfoBuilder.Build()
+	w.WriteHeader(http.StatusOK)
```
Lercerss: StatusOK is implicitly added when writing data to the response body as part of the `json.Encoder.Encode` call. Furthermore, if the encoding fails the StatusOK should not be written to the request's header, and writing the StatusOK after the encoding has succeeded results in a `superfluous write to headers` warning. I have removed these `WriteHeader` calls from other handlers in #152 for that reason and I think we should do the same here.

core/graphelier-service/api/hndlrs/orderhandler.go
```diff
@@ -0,0 +1,66 @@
...
+func getClosestSODTimestamp(timestamp uint64) (SODTimestamp int64) {
+	timeObject := time.Unix(0, int64(timestamp))
+	closestSOD := timeObject.Truncate(24 * time.Hour) //Rounds off to the closest day
+	return closestSOD.UnixNano()
+}
+
+// getClosestSODTimestamp returns the timestamp of the closest end of day
+func getClosestEODTimestamp(timestamp uint64) (SODTimestamp int64) {
+	timeObject := time.Unix(0, int64(timestamp))
+	closestSOD := timeObject.Truncate(24 * time.Hour) // Rounds off to the closest day
```
Lercerss: Replace this with a call to `getClosestSODTimestamp`?

core/graphelier-service/api/hndlrs/orderhandler_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"graphelier/core/graphelier-service/api/hndlrs"
+	. "graphelier/core/graphelier-service/utils/test_utils"
+)
+
+func TestFetchOrderInfoSuccess(t *testing.T) {
+
+	mockedDB := MockDb(t)
+	defer Ctrl.Finish()
+	messages := []*models.Message{&models.Message{Direction: -1, Instrument: "test", Type: models.NewOrder, OrderID: 1, Price: 100, ShareQuantity: 10, Timestamp: 1, SodOffset: 1}}
```
Lercerss: You should use the `MakeMsg` factory method here:
```suggestion
	messages := []*models.Message{MakeMsg(DirectionAsk)}
```
All the default values of the factory correspond to the values you entered here.

core/graphelier-service/api/hndlrs/orderhandler_test.go
```diff
@@ -0,0 +1,69 @@
...
+	mockedDB.EXPECT().
+		GetSingleOrderMessages("test", int64(1579564800000000000), int64(1579651200000000000), int64(1)).
+		Return(messages, nil)
+
+	var orderInfo models.OrderInfo
+	err := MakeRequest(
+		hndlrs.FetchOrderInfo, // Function under test
+		mockedDB,
+		"GET",
+		"/order/test/1/2",
```
Lercerss: It would be best to have the URL string match the `mux` params set on the request

core/graphelier-service/api/hndlrs/orderhandler_test.go
```diff
@@ -0,0 +1,69 @@
...
+		map[string]string{
+			"instrument": "test",
+			"orderID":    "1",
+			"timestamp":  "1579580821000000000",
+		},
+		&orderInfo,
+	)
+	assert.Nil(t, err)
+	assert.EqualValues(t, 100, orderInfo.Price)
```
Lercerss: Can you add a comment describing what you are asserting for here? To gain a higher level perspective than "price should be 100"

core/graphelier-service/models/orderinfo_test.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var newOrderMessage *Message = &Message{Direction: -1, Instrument: "test", Type: NewOrder, OrderID: 1, Price: 100, ShareQuantity: 10, Timestamp: 1, SodOffset: 1}
+var modifyMessage *Message = &Message{Direction: -1, Instrument: "test", Type: Modify, OrderID: 1, Price: 100, ShareQuantity: 2, Timestamp: 2, SodOffset: 2}
+var ignoreMessage *Message = &Message{Direction: -1, Instrument: "test", Type: Ignore, OrderID: 1, Price: 100, ShareQuantity: 10, Timestamp: 3, SodOffset: 3}
+var executeMessage *Message = &Message{Direction: -1, Instrument: "test", Type: Execute, OrderID: 1, Price: 100, ShareQuantity: 8, Timestamp: 4, SodOffset: 4}
```
Lercerss: The `MakeMsg` factory method would help distinguish the meaningful values between each message.

core/graphelier-service/models/orderinfo.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		orderInfo.LastModifiedTimestamp = message.Timestamp
+
+		if message.Type == NewOrder {
+			orderInfo.Quantity = message.ShareQuantity
+			orderInfo.Price = message.Price
+			orderInfo.CreatedOn = message.Timestamp
+		} else if message.Type == Modify || message.Type == Execute {
+			orderInfo.Quantity -= message.ShareQuantity
+		}
```
Lercerss: While it might not be a front-end use case, I still think this loop should handle `DELETE` messages and set the `orderInfo.Quantity = 0`.
```suggestion
		} else if message.Type == Delete {
			orderInfo.Quantity = 0
		}
```
I think it would make more sense to handle it than to return a non-zero quantity for deleted orders as-of a given timestamp.

core/graphelier-service/api/hndlrs/orderhandler_test.go
chimano: This test is mostly used to see if the the db gets called with the correct parameters these assertions are just there to confirm that the handler doesn't change anything

core/graphelier-service/api/hndlrs/orderhandler_test.go
chimano: Woops was a remnant of the copy paste

core/graphelier-service/api/hndlrs/orderhandler_test.go
Lercerss: Can you add a comment saying that in the file? To keep in line with #150

core/graphelier-service/api/hndlrs/orderhandler_test.go
```diff
@@ -0,0 +1,73 @@
...
+	mockedDB.EXPECT().
+		GetSingleOrderMessages("test", int64(1579564800000000000), int64(1579651200000000000), int64(1)).
+		Return(messages, nil)
+
+	var orderInfo models.OrderInfo
+	err := MakeRequest(
+		hndlrs.FetchOrderInfo, // Function under test
+		mockedDB,
+		"GET",
+		"/order/test/1/2",
```
Lercerss: Missed this one

core/graphelier-service/api/hndlrs/orderhandler.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// getClosestSODTimestamp returns the timestamp of the closest start of day
+func getClosestSODTimestamp(timestamp uint64) (SODTimestamp int64) {
+	timeObject := time.Unix(0, int64(timestamp))
+	closestSOD := timeObject.Truncate(24 * time.Hour) //Rounds off to the closest day
+	return closestSOD.UnixNano()
+}
+
+// getClosestSODTimestamp returns the timestamp of the closest end of day
```
alvyn279: comment: `getClosestEODTimestamp`

Issue #155, opened 2020-01-21
Change native DatePicker to @material-ui/pickers DatePicker


panoreak:
The native DatePicker used in the frontend should be replaced with the [@material-ui/pickers](https://material-ui-pickers.dev/demo/datepicker) DatePicker. The DatePicker  from [@material-ui/pickers](https://material-ui-pickers.dev/demo/datepicker) is much more configurable and user-friendly, and is the recommended input method in the [Material-UI docs](https://material-ui.com/components/pickers/).

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 1    |
| Priority     | Medium |
| Risk         | Low |

# Tasks
- Replace current DatePicker with the [@material-ui/pickers](https://material-ui-pickers.dev/demo/datepicker) DatePicker

Pull request: [#160]

panoreak:
### Acceptance test
| Steps                    | Inputs | Expected Outputs                                          | Status |
|--------------------------|--------|----------------------------------------------------------|--------|
| Visit the home page and visit the orderbook page | N/A | N/A | N/A   |
| Select and instrument and type 01/01/2020 in the date picker field |  N/A    | The orderbook should be displayed along with a graph depicting the best bid and best ask through the selected date | Passed |
| Select and instrument and pick 02/01/2020 using the date picker's popup selector| N/A  | The orderbook should be displayed along with a graph depicting the best bid and best ask through the selected date| Passed|

Issue #108, opened 2019-11-13
Create Top of Book graph component


panoreak:
Create a graph that shows the value of the best bid and best ask of the order book for an entire day (graph functionality will be extended in future tasks).

# Tasks
- [x] Make mockup
- [x] Find a suitable graph library that will allow zooming/panning
- [x] Create a graph component that shows best bid/ask price over time for the selected day
- [x] Load graph for selected day when a day is picked. Set selected time to 9:30 am by default
        -Make request using route from issue #111
        -Use data to populate graph
- [x] Reload order book when a time is clicked on the graph, put marker at that time (keep marker at selected time)

Story: [#86]
Pull request: [#160]

panoreak:
Quick Mockup of best asks/bids over time:
![image](https://user-images.githubusercontent.com/25068506/69189263-0e7f3a00-0aec-11ea-9dc3-4ddeb4d174a3.png)

panoreak:
> Quick Mockup of best asks/bids over time:
> ![image](https://user-images.githubusercontent.com/25068506/69189263-0e7f3a00-0aec-11ea-9dc3-4ddeb4d174a3.png)

Mockup was shown to Piere-Marc, he expressed concerns about using a graph library that isn't finance-specific. 

We need to use a graph that has the following features:
-Overlapping volume
-Discontinuous x-axis (if displaying several days, hide hours where market is closed, end of one day should be stuck to start of next day on x-axis)
-Zooming
-Showing bid/ask price at the cursor's position on the graph

Will look into [React Stockcharts](http://rrag.github.io/react-stockcharts/documentation.html#/overview) for creating the graph.

panoreak:
After looking into React Stockcharts we realized that we might need to create a Date wrapper. The library uses Date objects as the values for the X-axis. However, Date objects do not have nanosecond precision; we would add this precision with a wrapper class.

panoreak:
Comment from meeting with Pierre-Marc:
- Add x-axis labels

PR #158, opened 2020-01-25
[#108] Top of Book Graph, [#155] New Date Picker

panoreak:
Added TopOfBookGraph in the frontend [#108].

image

Features of the graph:
- Requests data points from backend whenever a new date is picked, or a new instrument is picked
- You can select a timestamp for viewing the state of the order book by clicking a time on the graph
- A message is shown instead of the graph if the backend request results in an error

Zoom functionality is still pending.

Also contains a new date picker [#155].


app/src/components/TopOfBookGraphWrapper.tsx
```diff
@@ -0,0 +1,30 @@
...
+            >
+                <TopOfBookGraph
+                    height={height}
+                    width={width}
+                    onTimeSelect={onTimeSelect}
+                    selectedDateTimeNano={selectedDateTimeNano}
+                    topOfBookItems={topOfBookItems}
+                />
+            </div>
+        </div>
```
chris-baur: Does this need two **div** ?

alvyn279:
@panoreak can we close this since #160 contains your changes ?

Issue #123, opened 2019-11-20
Add loading icon while loading new order book state


Lercerss:
As a user, I want to know the application is busy processing my request and not stuck.

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 1 |
| Priority     | Medium |
| Risk         | Low |

# Tasks
- Add loading icon when the user selects a new time

# Demo steps
- Select a different time
- While data is being processed, a loading icon is displayed
- Loading icon disappears once processing is done

## Customer signoff
- [ ] Pierre-Marc Fournier

[//]: # (Make sure to add the appropriate feature label to your issue)

alvyn279:
### Acceptance test
| Steps                    | Inputs | Expected Outputs                                          | Status |
|--------------------------|--------|----------------------------------------------------------|--------|
| Visit the home page and visit the orderbook page | N/A | Observe that a graphelier icon pops up with loading | PASS   |
| Select an instrument, and select January 1, 2020 as a day |  N/A    | The orderbook should be displayed along with a graph depicting the best bid and best ask through the selected date | PASS |
| Choose a time | Click on input method | Observe a new orderbook being displayed | PASS |
| Choose a different time | Click on input method | Observe the already mounted top of the book and message list to be greyed out during the load | PASS |

PR #159, opened 2020-01-26
[#123] Kontrolol/loading

chris-baur:
#123
- Added Greetings for loading App for the first time
- Added UI Feedback to show that data is loading

Issue
- Can not use desirable components for showing data being loaded because the entire UI freezes when performing the **Top of the Book** Logic, causing the circling or moving effects to also freeze.

### DIFF
--- a/app/src/components/TimestampOrderBookScroller.tsx
+++ b/app/src/components/TimestampOrderBookScroller.tsx
@@ -51,13 +51,10 @@ class TimestampOrderBookScroller extends Component<Props> {
     }
 
     componentDidUpdate(prevProps, prevState, snapshot) {
-        const { listItems, loading } = this.props;
+        const { listItems } = this.props;
 
         if (!listItemsEquals(prevProps.listItems || {}, listItems || {})) {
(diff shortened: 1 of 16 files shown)


app/src/components/TimestampOrderBookScroller.tsx
```diff
@@ -38,19 +40,25 @@ class TimestampOrderBookScroller extends Component<Props> {
...
 
     componentDidUpdate(prevProps, prevState, snapshot) {
-        const { listItems } = this.props;
+        const { listItems, loading } = this.props;
 
         if (!listItemsEquals(prevProps.listItems || {}, listItems || {})) {
             this.handleScrollToTopOfTheBook();
         }
+        if (loading !== prevProps.loading) {
+            this.forceUpdate();
```
alvyn279: Is this `forceUpdate()` necessary ? can you try without it, since a change in prop should re-render the component.

chris-baur: Did not remove it when finally pushing. Originally was testing something out and was not working as intended.

Issue #132, opened 2019-11-27
Go to most recent message in message list


Lercerss:
As a user, I want to be able to quickly go to the most recently applied message in the message list instead of scrolling manually until seeing the highlighted message.

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 3 |
| Priority     | High |
| Risk         | Low |

# Tasks
- Add button which scrolls to most recent message

# Demo steps
- Scroll the message list for a long time
- Return to most recently applied messages

## Customer signoff
- [ ] Pierre-Marc Fournier

[//]: # (Make sure to add the appropriate feature label to your issue)

chimano:
### Acceptance test
| Steps                    | Inputs | Expected Outputs                                          | Status |
|--------------------------|--------|----------------------------------------------------------|--------|
| Visit the home page and visit the orderbook page | N/A | N/A | N/A   |
| Select an instrument and select any time within January 1, 2020 |  N/A    | The orderbook should be displayed | Not yet tested  |
| Scroll through the message lists| N/A  | N/A | N/A|
| Using a button or a key go back to the most recent message applied on the orderbook| Press "`" on the keyboard  | The most recent message applied must be centered in the components | Not yet tested|

PR #161, opened 2020-01-26
[#132] Kontrolol/132/return to recent message

chris-baur:
#132
Recentering to the selected message now works after all intuitive manipulations of the visible orderbook
- New message selected
- Traversing messages with the **Forward** and **Backward** arrows
- selecting a new date and time
- Manually via the Tilde `

Issue
- Cannot determine properly to recenter when message selected within < 5 of an edge and **not** recenter when it is simply continuous scrolling from top or bottom.

### DIFF
--- a/app/src/components/MessageList.tsx
+++ b/app/src/components/MessageList.tsx
@@ -90,7 +90,8 @@ class MessageList extends Component<Props, State> {
                 this.fetchInitialMessages();
             }
             this.handleScrollBackToSelectedMessage();
-        } else if ((messages[0] && !prevState.messages[0])
+        } else if ((messages[0] && !prevState.messages[0]) // need to check for first time messages loaded
+            // need to check for new messages from new date and time, not from continuous scrolling
             || (messages[0] && prevState.messages[0]
                 && (messages[0].timestamp !== prevState.messages[0].timestamp
                     && messages[messages.length - 1].timestamp
--- a/app/src/components/MessageList.tsx
+++ b/app/src/components/MessageList.tsx
@@ -1,7 +1,8 @@
 /* eslint-disable camelcase */
 import React, { Component, createRef } from 'react';
 import { withStyles, createStyles, WithStyles } from '@material-ui/core/styles';
-import { Box } from '@material-ui/core';
+import { Box, Tooltip } from '@material-ui/core';
+import InfoIcon from '@material-ui/icons/Info';
 import classNames from 'classnames';
 import Button from '@material-ui/core/Button';
 import bigInt from 'big-integer';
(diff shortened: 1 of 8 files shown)


app/src/components/MessageList.tsx
```diff
@@ -124,6 +157,10 @@ class MessageList extends Component<Props, State> {
         }
     };
 
+    private onKeyUp = e => {
```
alvyn279: jsdoc :)

app/src/components/MessageList.tsx
```diff
@@ -221,6 +263,9 @@ class MessageList extends Component<Props, State> {
             <div className={classes.scrollContainer}>
                 <Box className={classes.tableHeaderRow}>
                     <Box className={classNames(classes.tableColumn, classes.overrideTimestampColumn)}>
+                        <CustomTooltip title={'Click the Tilde ` to manually center to selected message'}>
```
alvyn279: "Click tilde ` to center to selected message"

app/src/components/MessageList.tsx
```diff
@@ -221,6 +263,9 @@ class MessageList extends Component<Props, State> {
...
+                            <InfoIcon />
+                        </CustomTooltip>
```
alvyn279: can this custom tooltip have a space after it so that it's a little separated from "Timestamp" label

app/src/components/MessageList.tsx
chris-baur: 👍

app/src/components/MessageList.tsx
chris-baur: 👍

app/src/components/MessageList.tsx
chris-baur: 👍

app/src/components/MessageList.tsx
```diff
@@ -77,9 +87,20 @@ class MessageList extends Component<Props, State> {
             } else {
                 this.fetchInitialMessages();
             }
+            this.handleScrollBackToSelectedMessage();
+        } else if ((messages[0] && !prevState.messages[0])
+            || (messages[0] && prevState.messages[0]
+                && (messages[0].timestamp !== prevState.messages[0].timestamp
+                    && messages[messages.length - 1].timestamp
+                    !== prevState.messages[prevState.messages.length - 1].timestamp))) {
```
chimano: Is there a way to simplify this?

chimano: If not can you write a comment on what it's checking for?

chris-baur: sure, I'll write it out here and make sure it makes sense, and is what we want.

chris-baur: 1. Need to check for messages being given for the first time, where prevState would have an emprty array and current state is populated ( messages[0] && !prevState.messages[0] )
2. Need to make sure that when there is a new selected date and time (Currently the slider, but later graph would determine it) that it recenters the new selected message. We also need to make sure that it does not do this centering when it is simply new messages being added to the list of messages via the continuous scrolling.

alvyn279: i think he means in code comment, which would require more concise explanation

Issue #170, opened 2020-01-31
Scale on top of book graph is off by an hour


alvyn279:
The scale for the top of book graph shows that there is data from 8:30AM to 3:00PM. This is not business hours.

This problem might be linked with daylight saving time at the instantiation of `Date` objects.

| Metric       | Value           |
|--------------|-----------------|
| Priority     | High |
| Risk         | Low |

# Tasks
- Instantiate the `Date` objects used by the graph with the correct timezone offset (EDT)

Lercerss:
It appears this was actually not fixed as it re-occurs now, after daylight savings changed again.

alvyn279:
fixed in #202

PR #171, opened 2020-01-31
[#170] Fixed scale being offset by an hour

alvyn279:
Scale now displays time from 9:30 AM to 4:00 PM.


alvyn279:
> Is there a reason that we now want to show how we measured the constant? I.e showing the formula instead of just putting the product.
> 
> Looks good

yea makes it more verbose as to how we got to that number

chris-baur:
> > Is there a reason that we now want to show how we measured the constant? I.e showing the formula instead of just putting the product.
> > Looks good
> 
> yea makes it more verbose as to how we got to that number

Ok, 👍

Lercerss:
Actually, why are we subtracting one hour? Won't this fail when daylight savings changes again?

alvyn279:
> Actually, why are we adding one hour? Won't this fail when daylight savings changes again?

Because Date object internally changes the given utc timestamp into local timezone, which involves daylight saving time. Until we're not going to interact with JS' Date interface, we'll have to change it every time dst changes, since we agreed for constant EDT timezone. 

Correct me if i'm not seeing a better solution

Lercerss:
Have you looked into ways of instantiating a Date object with a given timezone instead of defaulting to the local one?

Lercerss:
If that is not possible, I would prefer that we add/subtract the difference between the local timezone and EDT at runtime instead of hardcoding it to a value

alvyn279:
reading this: https://stackoverflow.com/a/15171030

Issue #172, opened 2020-02-02
Graph does not display points for a full day


Lercerss:
The left bound of the graph is not always 9:30:00 AM, rather sometimes it is ~9:56:00 AM.
This has been observed using Chrome on a 1080p display with data from MSFT on the production server.

Additionally, zooming using the browser also results in the left bound to move. Zooming out allows displaying time in the past for which there are no points loaded. Zooming in removes points from the left.

Pull requests: [#173] [#174]

Lercerss:
Re-opening to fix the zooming in/out on the browser after more investigation by @panoreak

PR #173, opened 2020-02-03
[#172] Bug 172

panoreak:
Fixed bugs mentioned in issue #172:

- The graph now shows a full day from 9:30 am to 4:00 pm
- Data points are left intact after zooming in/out with the browse or after resizing the window

The bug fix was tested with all three instruments, using Chrome on a 1080p display, with the backend server.

Also fixed the bug where the message "Could not retrieve graph for this day." is shown while the graph is loading. It has been replaced with a spinner, and the message is only displayed if the request for data points returns an error.


app/src/components/OrderBookSnapshot.tsx
```diff
@@ -116,7 +120,7 @@ class OrderBookSnapshot extends Component<WithStyles, State> {
      * @returns {number}
      */
     getNumDataPoints = (): number => {
-        return Math.trunc(window.screen.width * window.devicePixelRatio * 0.76);
+        return Math.trunc(window.screen.width * window.devicePixelRatio * 0.5);
```
Lercerss: Since there is a Constants file, shouldn't we place this one there?

app/src/components/OrderBookSnapshot.tsx
```diff
@@ -396,24 +412,31 @@ class OrderBookSnapshot extends Component<WithStyles, State> {
...
+                                    {loadingGraph && (
+                                        <div className={classes.graphLoader}>
+                                            <CustomLoader
+                                                size={'5rem'}
+                                                type={'circular'}
                                             />
-                                        )}
+                                        </div>
+                                    )}
+                                    { !loadingGraph && !graphUnavailable && topOfBookItems.length !== 0 && (
```
Lercerss: Is there ever a case where `topOfBookItems.length` is 0? If so, should we handle it?

app/src/components/TopOfBookGraphWrapper.tsx
```diff
@@ -1,28 +1,87 @@
...
+            }, () => this.render());
+        }
+    };
+
+    render() {
+        const {
+            classes, onTimeSelect, selectedDateTimeNano, topOfBookItems,
+        } = this.props;
+        const { graphWidth, graphHeight } = this.state;
+        console.log('dimensions', graphWidth, graphHeight);
```
Lercerss: Can you drop this log before we hit production? It seems a bit spammy while testing it out locally. Otherwise if we want to keep it I'd like to make it prettier

panoreak: I'll drop it

app/src/components/OrderBookSnapshot.tsx
panoreak: Hmm, I don't think so, unless we somehow request 0 data points. I can remove that condition.

panoreak: Wait, actually, the initial state of OrderBookSnapshot has the topOfBookItems set to empty

Lercerss: So I guess we simply don't want to display anything in that case?

panoreak: Yeah, we don't want to display anything. If you give an empty array of data to the chart library, it throws an error, so I think it's safest to keep this condition check

app/src/components/OrderBookSnapshot.tsx
panoreak: You're right, I'll move it

panoreak:
> See my other comments about the code.
> 
> Trying it out locally, it seems to work for the default case with all instruments. However, zooming in/out with the browser still messes up the graph, though I am not sure to which extent we want to support this. I also noticed that having the developer console open still skewed the points on the left hand side by a bit

I know what you're talking about. One thing to consider is that once zooming/panning will be supported, you'll be able to scroll tot he left or zoom out to see the earlier data points that get hidden when the window is resized.

app/src/components/TopOfBookGraphWrapper.tsx
```diff
@@ -1,28 +1,86 @@
...
+}
+
+interface State {
+    graphWidth: number,
+    graphHeight: number,
+}
+
+
+class TopOfBookGraphWrapper extends Component<Props, State> {
+    private readonly divRef: React.RefObject<HTMLDivElement>;
```
alvyn279: can this div ref have a more significant name :)

app/src/components/OrderBookSnapshot.tsx
```diff
@@ -396,24 +412,31 @@ class OrderBookSnapshot extends Component<WithStyles, State> {
...
+                                            className={classes.noDataMessage}
+                                            variant={'body1'}
+                                            color={'textPrimary'}
+                                        >
+                                            {'Could not retrieve graph for this day.'}
+                                        </Typography>
+                                    )}
+                                    {loadingGraph && (
+                                        <div className={classes.graphLoader}>
+                                            <CustomLoader
```
alvyn279: :+1:

PR #174, opened 2020-02-05
[#172] Fixed issue with X-axis on window resize

panoreak:
Fixed issue where X-axis is truncated on the left when the window is resized, and no longer shows 9:30 AM.

Changes:
- Added call to `updateGraphData` when window is resized
- Number of data points requested now depends on window width rather than screen width

### DIFF
--- a/app/src/components/OrderBookSnapshot.tsx
+++ b/app/src/components/OrderBookSnapshot.tsx
@@ -9,6 +9,7 @@ import { WithStyles, createStyles } from '@material-ui/core/styles';
 import MomentUtils from '@date-io/moment';
 import { MuiPickersUtilsProvider, KeyboardDatePicker } from '@material-ui/pickers';
 import bigInt from 'big-integer';
+import { debounce } from 'lodash';
 
 import moment from 'moment';
 import { Styles } from '../styles/OrderBookSnapshot';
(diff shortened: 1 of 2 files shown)


app/src/components/OrderBookSnapshot.tsx
```diff
@@ -94,10 +97,26 @@ class OrderBookSnapshot extends Component<WithStyles, State> {
         }
     }
 
+    componentWillUnmount() {
+        window.removeEventListener('resize', debounce(this.handleResize, 100));
```
Lercerss: Using a REPL, I have found that two calls to `debounce` with the same arguments do not match with `==` or `===`, therefore I am not sure if the event listener will really be removed. We could avoid this by simply storing the debounced function as an attribute on the OrderBookSnapshot object

app/src/components/OrderBookSnapshot.tsx
```diff
@@ -51,6 +52,18 @@ interface State {
...
+     * @desc Handles window resizing and requests a new number of data points appropriate for the new window width
+     */
+    handleResize = debounce(() => {
+        const { selectedDateNano } = this.state;
+        if (selectedDateNano.neq(0)) {
+            const startTime = selectedDateNano.plus(NANOSECONDS_IN_NINE_AND_A_HALF_HOURS);
+            const endTime = selectedDateNano.plus(NANOSECONDS_IN_SIXTEEN_HOURS);
+            this.updateGraphData(startTime, endTime);
+        }
+    }, 100);
```
alvyn279: is there a reason we wait a certain amount of time before sending the request again

panoreak: While dragging the browser window to resize it/while dragging the dev tools side bar, many requests are sent to the backend. I found 0.1 seconds to be a reasonable waiting time to minimize the number of requests being sent.

Issue #164, opened 2020-01-26
Add messages to small interval


pvacca97:
The backend will need to service the case where it is required to send more points than snapshot intervals to the frontend. This requires adding enough messages, spaced out evenly, within the range requested.

Related story: #165 

# Tasks
- Add the small interval case to the existing route `/topbook/{instrument}/{start_timestamp}/{end_timestamp}/{num_points}`
- Get the latest orderbook from the lowwer bound of the interval
- Apply messages evenly until the higher bound of the interval
- Find top of book at each of these points and return the result

panoreak:
I was thinking about the cases where we're requesting very precise data points, and I realized that if the number of nanoseconds between the requested start time and end time is smaller than the number of requested data points, you should limit the number of data points to the number of nanoseconds between that interval.

To put it a different way, you should make sure that there is max one data point per nanosecond timestamp that you return, even if that means that you are returning less data points than the number of data points requested.

PR #183, opened 2020-02-10
[#164] zoom small interval api

Lercerss:
adds api functionality to zoom in to smaller intervals for the graph


core/graphelier-service/models/orderbooks.go
```diff
@@ -267,3 +267,24 @@ func (orderbook *Orderbook) ApplyMessagesToDeltabook(messages []*Message, numMes
...
 }
+
+// TopBookPerXNano : Returns a list of the top book at every x nanosecond after applying each individual message in a list
+func (orderbook *Orderbook) TopBookPerXNano(messages []*Message, x int) (topbook []*Point) {
+	firstNano := messages[0].Timestamp
+	lastNano := messages[len(messages)-x].Timestamp
+
+	for counter, nano := uint64(0), firstNano; nano <= lastNano; counter, nano = counter+1, nano+1 {
+		nanoMessages := GetMessagesForTimestamp(messages, nano)
+		orderbook.ApplyMessagesToOrderbook(nanoMessages)
```
chimano: I think you should be iterating over messages rather than by nanoseconds since there'll be many nanoseconds where there will be no corresponding message

core/graphelier-service/models/messages.go
```diff
@@ -35,3 +35,16 @@ type MessagePage struct {
 	PageInfo Paginator  `json:"pageInfo"`
 	Messages []*Message `json:"messages"`
 }
+
+// GetMessagesForTimestamp : Returns all messages for a given nanosecond
+func GetMessagesForTimestamp(messages []*Message, nano uint64) (messagesFromTimestamp []*Message) {
```
chimano: If you iterate over messages you won't need this helper function

core/graphelier-service/db/mongo.go
```diff
@@ -336,52 +337,110 @@ func (c *Connector) GetTopOfBookByInterval(instrument string, startTimestamp uin
 	}
 	log.Tracef("Documents in interval {%d,%d}=%d, keeping 1 in %d", startTimestamp/interval, endTimestamp/interval, count, count/maxCount)
 
-	// Add $mod comparator to interval_multiple filter
-	filter[1].Value = append(filter[1].Value.(bson.D), bson.E{Key: "$mod", Value: bson.A{count / maxCount, 0}})
+	if maxCount < count {
```
chimano: The function is getting pretty massive, can it be split up?

core/graphelier-service/db/mongo.go
```diff
@@ -22,6 +22,7 @@ type Datastore interface {
 	RefreshCache() error
 	GetSingleOrderMessages(instrument string, SODTimestamp int64, EODTimestamp int64, orderID int64) ([]*models.Message, error)
 	GetTopOfBookByInterval(instrument string, startTimestamp uint64, endTimestamp uint64, maxCount int64) (results []*models.Point, err error)
+	GetMessagesWithinInterval(instrument string, startTimestamp uint64, endTimestamp uint64) (results []*models.Message, err error)
```
chimano: From what i remember this can return more than an interval's worth of messages so it might need to be renamed. @Lercerss can you confirm?

core/graphelier-service/db/mongo.go
```diff
@@ -336,52 +337,110 @@ func (c *Connector) GetTopOfBookByInterval(instrument string, startTimestamp uin
...
+		}
+		if int64(exactEnd-exactStart) <= maxCount {
+			// return every nanosecond, not enough points
+			// x = 1 for no spacing between points
+			results = orderbook.TopBookPerXNano(messages, 1)
+		} else {
+			// return spaced nanoseconds, too many points
+			// do not care about truncating
+			x := len(messages) / int(maxCount)
+			results = orderbook.TopBookPerXNano(messages, x)
```
chimano:
```
x := len(messages) / int(maxCount)
if x == 0 {
  x = 1
}
results = orderbook.TopBookPerXNano(messages, x)
```

core/graphelier-service/models/orderbooks.go
alvyn279: @chimano will this be possible if we want uniform distance between each point given back

chimano: Definitely, the state of the orderbook doesn't change between messages so if we have m1 at 97s and m2 at 105s and we want points at a 100s interval, m1 will be applied, once m2 is reached we can see that it's at another multiple, so then the best bid/ask will be stored at 100s and then it'll continue applying the rest of the messages for the other times

pvacca97: i did this because there will sometimes be multiple messages at a nanosecond that need to be applied together

core/graphelier-service/db/mongo.go
Lercerss: I actually have something very similar in #184 with a [different name](https://github.com/Lercerss/graphelier/pull/184/files#diff-1717a8460c15d66b9eab79dbc9668be6R19)

core/graphelier-service/models/orderbooks.go
chimano: with some extra logic you can still do that by iterating messages

core/graphelier-service/db/mongo.go
pvacca97: it's responsible for the large and small intervals depending on the number of data points we receive from the frontend. if you want we can pass the cursor into another function to just get the results for the large interval?

chimano: Yeah that works

core/graphelier-service/models/orderbooks.go
```diff
@@ -267,3 +267,31 @@ func (orderbook *Orderbook) ApplyMessagesToDeltabook(messages []*Message, numMes
...
 }
+
+// TopBookPerXNano : Returns a list of the top book at every x nanosecond after applying each individual message in a list
+func (orderbook *Orderbook) TopBookPerXNano(messages []*Message, x uint64) (topbook []*Point) {
+	topbookLengthIndex := 0
+	initialTimestamp := messages[0].Timestamp
+	for _, message := range messages {
+		singleMsgSlice := []*Message{message}
+		orderbook.ApplyMessagesToOrderbook(singleMsgSlice)
+		if (message.Timestamp-initialTimestamp)%x == 0 {
```
chimano: This will skip over a few points.
Ex:
message nanos: [1 2 5 6]
You'll only record the point at 6 when there should be a point at 3 and 6

core/graphelier-service/models/orderbooks.go
```diff
@@ -267,3 +267,31 @@ func (orderbook *Orderbook) ApplyMessagesToDeltabook(messages []*Message, numMes
...
+			}
+			if len(orderbook.Asks) != 0 {
+				tb.BestAsk = orderbook.Asks[0].Price
+			}
+			if topbookLengthIndex > 0 && message.Timestamp == topbook[topbookLengthIndex-1].Timestamp {
+				topbook[topbookLengthIndex-1] = tb
+			} else {
+				topbook = append(topbook, tb)
+				topbookLengthIndex++
+			}
```
chimano: What is this for?

pvacca97: it is my attempt to replace the best bid/ask if we have multiple messages at the same timestamp

core/graphelier-service/models/orderbooks.go
pvacca97: I see... I'm not sure i understand your solution mentioned previously, can you explain again please?

core/graphelier-service/models/orderbooks.go
```diff
@@ -267,3 +267,29 @@ func (orderbook *Orderbook) ApplyMessagesToDeltabook(messages []*Message, numMes
 
 	return
 }
+
+// TopBookPerXNano : Returns a list of the top book at every x nanosecond after applying each individual message in a list
+func (orderbook *Orderbook) TopBookPerXNano(messages []*Message, x uint64) (topbook []*Point) {
+	topbookLengthIndex := uint64(0)
+	for messagesIndex, message := range messages {
```
Lercerss: As @chimano had mentioned earlier, you are going to need more logic to make sure you are creating points at regular intervals and not just when there is a message. For example you do not generate points between the start of the interval and the first message, between the last message and the end of the interval, and you are only generating at most 1 point between 2 messages while there could be many.

core/graphelier-service/db/mongo.go
```diff
@@ -334,54 +385,69 @@ func (c *Connector) GetTopOfBookByInterval(instrument string, startTimestamp uin
...
+		}
+		messages, err := c.GetMessagesWithinInterval(instrument, startTimestamp, endTimestamp)
+		if err != nil {
+			return nil, err
+		}
+		x := len(messages) / int(maxCount)
+		if x == 0 {
+			x = 1
+		}
+		log.Tracef("Documents in message interval {%d,%d}=%d, keeping 1 in %d", startTimestamp, endTimestamp, len(messages), x)
```
Lercerss: The end of this log does not really make sense, we should not be "keeping 1 in %d" messages as they are all important.

core/graphelier-service/db/mongo.go
```diff
@@ -334,54 +385,69 @@ func (c *Connector) GetTopOfBookByInterval(instrument string, startTimestamp uin
 	if err != nil {
 		return nil, err
 	}
-	log.Tracef("Documents in interval {%d,%d}=%d, keeping 1 in %d", startTimestamp/interval, endTimestamp/interval, count, count/maxCount)
+	log.Tracef("Documents in orderbook interval {%d,%d}=%d, keeping 1 in %d", startTimestamp/interval, endTimestamp/interval, count, count/maxCount)
```
Lercerss: You should move this log inside the if statement

core/graphelier-service/utils/utils_test.go
```diff
@@ -0,0 +1,24 @@
+package utils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExistsMultiple(t *testing.T) {
```
Lercerss: Could you add asserts for the value of the boolean?

core/graphelier-service/db/mongo.go
```diff
@@ -334,54 +385,71 @@ func (c *Connector) GetTopOfBookByInterval(instrument string, startTimestamp uin
...
+	} else {
+		orderbook, err := c.GetOrderbook(instrument, startTimestamp)
+		if err != nil {
+			return nil, err
+		}
+		messages, err := c.GetMessagesWithinInterval(instrument, startTimestamp, endTimestamp)
+		if err != nil {
+			return nil, err
+		}
+		x := len(messages) / int(maxCount)
```
Lercerss: Shouldn't you be using the timestamp range here to determine how long your interval is?
```suggestion
		x := (endTimestamp - startTimestamp) / maxCount
```

core/graphelier-service/models/orderbooks.go
```diff
@@ -267,3 +267,30 @@ func (orderbook *Orderbook) ApplyMessagesToDeltabook(messages []*Message, numMes
...
+		orderbook.ApplyMessagesToOrderbook(singleMsgSlice)
+	}
+	// handles points after messages
+	topbook = orderbook.backfillPoints(topbook, pointDistance, lastMultipleCount, endTimestamp/pointDistance)
+
+	return topbook
+}
+
+// add points for previous multiples of point distance
+func (orderbook *Orderbook) backfillPoints(topbook []*Point, pointDistance uint64, leftMultiple uint64, rightMultiple uint64) []*Point {
```
chimano: :+1:

PR #186, opened 2020-02-14
Graph zooming & panning 

alvyn279:
Two new behaviours:

- When zooming, new data points are loaded onto the graph
- Data points are held with `NanoDate` objects, which conserves nanoseconds precision. Zooming can therefore be performed up until latter.


alvyn279:
Some other reviewers on this would be great

Issue #126, opened 2019-11-20
Create Order Information Component


riyaGit:
Create a " swipeable temporary drawer" component that shows in addition information for the order when an order is clicked in the order book. 

# Tasks
- [x] Create Mockup
- [x] Create react component that will contain order information
        The component will be a drawer component containing a table
- [x] Get data from the backend containing the additional order information
- [x] Change the colour of the order "box" when it is clicked on

Story: [#110 ]

riyaGit:
Mockup
<img width="1187" alt="Screen Shot 2020-01-16 at 2 42 22 PM" src="https://user-images.githubusercontent.com/26652806/72557464-ce7bcf80-386e-11ea-97b5-30462526566e.png">

Lercerss:
Seeing this mockup, it might be better if the order information was a more temporary component such as a swipeable drawer.

riyaGit:
I agree! This new mockup displays as such
<img width="1243" alt="Screen Shot 2020-01-16 at 3 05 13 PM" src="https://user-images.githubusercontent.com/26652806/72558946-b48fbc00-3871-11ea-8709-1503662c2af5.png">

chimano:
Just to clarify, orders are created, modified, executed, and cancelled on the same day. So we should be showing the datetime and not just the date.
Also to clarify, will the message list contain all messages that have modified the order and the ones that will modify it in the future or will it only contain all the messages that have modified it up until the requested timestamp?

Lercerss:
I believe it would be better to display all messages that affect the order, not just the ones that have been applied up until the requested timestamp. Maybe we could style the future messages differently for clarity?

riyaGit:
Resolved in PR #187

PR #187, opened 2020-02-16
Order information component #126

riyaGit:
Issue #126 

This is demo of what the order information looks like
Screen Recording 2020-02-16 at 1 41 25 PM

Things left to do that require redux:
1. Change colour of order component when clicked on
2. Change state of orderbook when messages are clicked on

### DIFF
--- a/app/src/components/Order.tsx
+++ b/app/src/components/Order.tsx
@@ -10,10 +10,11 @@ import bigInt from 'big-integer';
 import { connect } from 'react-redux';
 import { Styles } from '../styles/Order';
 import {
-    Message, OrderInformationDrawer, TransactionType, OrderDetails,
+    OrderInformationDrawer, TransactionType, OrderDetails,
 } from '../models/OrderBook';
 import OrderBookService from '../services/OrderBookService';
 import { showOrderInfoDrawer } from '../actions/actions';
+import { RootState } from '../store';
 
 const styles = createStyles(Styles);
 
@@ -25,17 +26,11 @@ interface Props extends WithStyles<typeof styles> {
     orderId: number,
     timestamp: bigInt.BigInteger,
     onOrderClicked: Function,
+    orderIdFromState: number,
+    orderInfoDrawerShown: boolean,
 }
 
-interface State {
-    lastModified: string,
-    createdOn: string,
-    price: number,
-    messages: Array<Message>,
-    infoDrawerShown : boolean,
-}
-
-class Order extends Component<Props, State> {
+class Order extends Component<Props> {
     /**
      * @desc handles event for sending request to retrieve order information when order is clicked on
      */
@@ -68,7 +63,7 @@ class Order extends Component<Props, State> {
 
     render() {
         const {
-            classes, type, quantity, maxQuantity,
+            classes, type, quantity, maxQuantity, orderId, orderIdFromState, orderInfoDrawerShown,
         } = this.props;
         const quantityBoxSize = (quantity / maxQuantity) * 100;
         const minQuantityTextSize = 2;
@@ -76,6 +71,11 @@ class Order extends Component<Props, State> {
             classes.quantity,
             classes.rectangle,
             type === TransactionType.Bid ? classes.bid : classes.ask,
+        );
+        const clickedOrderClass = classNames(
+            classes.quantity,
+            classes.rectangle,
+            classes.getOrderInfo,
         );
         const shouldShowQuantity = quantityBoxSize > minQuantityTextSize;
         return (
@@ -86,18 +86,28 @@ class Order extends Component<Props, State> {
                 classes={{ tooltip: classes.offsetTooltip }}
             >
                 <Box
-                    className={orderClasses}
+                    className={orderInfoDrawerShown
+                    && orderIdFromState === orderId ? clickedOrderClass : orderClasses}
                     style={{ width: `${quantityBoxSize}%` }}
                     onClick={this.handleOnOrderClick}
                 >
-                    {shouldShowQuantity && <Typography className={classes.text}>{quantity}</Typography>}
+                    {shouldShowQuantity && (
+                        <Typography className={orderInfoDrawerShown
+                    && orderIdFromState === orderId ? classes.orderTextOnClick : classes.text}
+                        >
+                            {quantity}
+                        </Typography>
+                    )}
                 </Box>
             </Tooltip>
         );
     }
 }
 
-const mapStateToProps = () => ({});
+const mapStateToProps = (state: RootState) => ({
+    orderInfoDrawerShown: state.general.showOrderInfoDrawer,
+    orderIdFromState: state.general.orderDetails ? state.general.orderDetails.id : -1,
+});
 
 const mapDispatchToProps = (dispatch : Dispatch) => ({
     // eslint-disable-next-line max-len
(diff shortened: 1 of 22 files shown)


app/src/components/Order.tsx
```diff
@@ -2,24 +2,89 @@ import React, { Component } from 'react';
...
+        this.state = {
+            infoDrawerShown: false,
+            lastModified: '',
+            createdOn: '',
+            price: 0,
+            messages: [],
+        };
+    }
+
+    handleOnOrderClick = () => {
```
alvyn279: jsdoc on every function

app/src/components/OrderInformation.tsx
```diff
@@ -0,0 +1,153 @@
...
+class OrderInformation extends Component<Props, State> {
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            right: true,
+        };
+    }
+
+    renderToggleDrawer(side, open) {
```
alvyn279: jsdoc

app/src/components/Order.tsx
```diff
@@ -2,24 +2,89 @@ import React, { Component } from 'react';
 import { withStyles, WithStyles, createStyles } from '@material-ui/core/styles';
 import classNames from 'classnames';
 
-import { Typography, Box, Tooltip } from '@material-ui/core';
+import {
+    // eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
```
riyaGit: Will remove this!

app/src/components/OrderInformation.tsx
```diff
@@ -0,0 +1,153 @@
...
+                return;
+            }
+
+            this.setState({ right: true, [side]: open });
+        };
+    }
+
+    renderSlideList(side) {
+        const {
+            // eslint-disable-next-line no-unused-vars,@typescript-eslint/no-unused-vars
```
riyaGit: this as well!

app/src/components/OrderInformation.tsx
```diff
@@ -0,0 +1,153 @@
...
+        return event => {
+            if (event && event.type === 'keydown' && (event.key === 'Tab' || event.key === 'Shift')) {
+                return;
+            }
+
+            this.setState({ right: true, [side]: open });
+        };
+    }
+
+    renderSlideList(side) {
```
alvyn279: this custom render function already breaks down a part of the table, which is good. Can it be broken down even more, as it is still big.

rendering of the tablebody maybe in another function ?

app/src/components/OrderInformation.tsx
```diff
@@ -0,0 +1,153 @@
...
+                            {'ID: '}
+                            {orderId}
+                        </h2>
+                    </TableHead>
+                    <TableRow>
+                        <TableCell>
+                            {'Quantity: '}
+                            {quantity}
+                        </TableCell>
+                    </TableRow>
```
alvyn279: image

can all of these info be right-aligned on the same line as the descriptor ? what do you think?

app/src/constants/Constants.ts
```diff
@@ -3,7 +3,7 @@ import bigInt from 'big-integer';
 export const ENVIRONMENT = 'DEV';
 export const APP_NAME = 'Graphelier';
 export const BACKEND_PORT = 5050;
-let host = 'localhost';
+let host = '18.218.121.174';
```
alvyn279: remove this

app/src/components/Order.tsx
alvyn279: except lifecycle methods and render ***

app/src/constants/Constants.ts
riyaGit: lol ya i did

app/src/components/Order.tsx
riyaGit: ok

app/src/components/OrderInformation.tsx
riyaGit: ya i think putting the message list table in the another function would be feasible!

app/src/components/OrderInformation.tsx
riyaGit: Good idea! i think that would look good

riyaGit: This is what it looks like when i right align values. Lmk what you think and i will commit
Screen Shot 2020-02-16 at 5 12 20 PM

riyaGit: Here is the differently aligned one as requested! 
Screen Shot 2020-02-16 at 7 53 37 PM

alvyn279:
Some clicks don't seem to work on some orders. I think it might be a thing with unique keys, as seen from the picture below.

image

app/src/components/Order.tsx
```diff
@@ -36,14 +86,32 @@ class Order extends Component<Props> {
...
 }
 
-export default withStyles(styles)(Order);
+const mapStateToProps = (state: RootState) => ({
+    orderInfoDrawerShown: state.general.showOrderInfoDrawer,
+    orderIdFromState: state.general.orderDetails ? state.general.orderDetails.id : -1,
+});
+
+const mapDispatchToProps = (dispatch : Dispatch) => ({
+    // eslint-disable-next-line max-len
```
alvyn279: just break the line instead of suppressing it

app/src/components/OrderInformation.tsx
```diff
@@ -0,0 +1,218 @@
...
+                </SwipeableDrawer>
+            </div>
+        );
+    }
+}
+
+const mapStateToProps = () => ({});
+
+const mapDispatchToProps = (dispatch : Dispatch) => ({
+    // eslint-disable-next-line max-len
```
alvyn279: not needed

app/src/components/OrderInformation.tsx
```diff
@@ -0,0 +1,218 @@
...
+                        } = message;
+
+                        const { timeNanoseconds } = splitNanosecondEpochTimestamp(bigInt(timestamp));
+                        const time = nanosecondsToString(
+                            convertNanosecondsUTCToCurrentTimezone(bigInt(timeNanoseconds)).valueOf(),
+                        );
+
+                        return (
+                            <TableRow
+                                key={order_id}
```
chris-baur: Won't this cause an issue since messages will have the same order id ?

app/src/components/OrderInformation.tsx
riyaGit: done

app/src/components/Order.tsx
riyaGit: done

app/src/components/OrderInformation.tsx
riyaGit: order ids are unique i believe
@chimano can you confirm?

Lercerss: Your keys won't be unique, all messages will have the same order_id as they all relate to the same order. You could use the sod_offset which uniquely identifies each message.

riyaGit: Perfect thanks Lucas, I'll do that then! Good catch Chris!

riyaGit:
> Some clicks don't seem to work on some orders. I think it might be a thing with unique keys, as seen from the picture below.
> 
> image

That's been fixed by changing the key of messaged from order_id to sod_offset!

panoreak:
lgtm

Issue #176, opened 2020-02-05
Animate modifications to the order book


Lercerss:
From the app perspective, we need to implement animations for modifications to the order book.

# Tasks
- Create mockup/prototype of animations for each of the possible modification types:
  - Delete an order
  - Add a new order
  - Modify an order (execute could be a different animation or the same as modify)
  - (Tentative) Move an order across price-levels
- Create component to apply animations on an order book state given a list of modifications

Story #175.

chimano:
New Order:
![NEW_ORDER](https://user-images.githubusercontent.com/26588048/74108124-ff38e700-4b44-11ea-8ef1-57bbd9969be3.gif)

Delete Order:
![DELETE_ORDER](https://user-images.githubusercontent.com/26588048/74108130-0bbd3f80-4b45-11ea-86a0-07fd99460220.gif)

chimano:
Modifying/Executing order:
![MODIFY_ORDER](https://user-images.githubusercontent.com/26588048/74110465-dae80500-4b5a-11ea-81d1-ff1d1036ff1b.gif)

chimano:
New Price Level:
![Add_Price](https://user-images.githubusercontent.com/26588048/75292096-877be500-57f1-11ea-8612-0bfb9f77890d.gif)
Delete Price Level:
![Remove_Price](https://user-images.githubusercontent.com/26588048/75292111-8f3b8980-57f1-11ea-9fe4-502e483f4ac9.gif)

PR #189, opened 2020-02-25
[#176] Added animations on pricelevels

chimano:
- Smooth transitions added for adding and removing pricelevels

Can see mockups at #176


app/src/components/TimestampOrderBookScroller.tsx
```diff
@@ -183,23 +185,32 @@ class TimestampOrderBookScroller extends Component<Props> {
...
-                                            // @ts-ignore
-                                            ref={listItem.isMiddle ? this.middleReferenceItem : null}
-                                            className={classes.pricePoint}
+                                            timeout={ANIMATION_TIME}
+                                            classNames={{
+                                                enter: classes.levelEnter,
+                                                enterActive: classes.levelEnterActive,
+                                                exit: classes.levelExit,
+                                                exitActive: classes.levelExitActive,
+                                            }}
```
alvyn279: `classNames` (and not `className`) was possible ?

chimano: `classNames` is actually a special prop for `CSSTransition` that specifies the css class the element will have for each transition phase (`enter`, `enterActive`, `exitActive`, `exit`). Can't be replaced by the basic `className`.

Issue #191, opened 2020-02-27
Unable to zoom out back to a full day's view


Lercerss:
After zooming in a lot (interval become less than 5 minutes long), the graph is unable to zoom out fully again. Instead, it prevents zooming out further than an interval around 25 minutes. Clicking on the graph, panning or zooming in and out again do not seem to resolve the issue. Inspecting the console logs, it seems the graph still sends a new request on scroll but the bounds don't change.

Lercerss:
On a separate note, it seems sometimes the graph does not load more points when panning. Exact steps to reproduce it are unclear, but once it starts occurring the only way to load more points is to zoom the graph.

alvyn279:
Two things to note:

- Zooming with a mouse wheel on Chrome works fine it seems, only using touchpad zoom in/out leads to the whole page being dragged
- There indeed seems to be an issue with panning not loading more points, although the steps to reproduce seem to be inconsistent.

PR #194, opened 2020-02-29
[#191] Fixed zoom out bug when zoomed in too deep

alvyn279:
Zooming out once zoomed in too deep is now unblocked.

zoom_graph


alvyn279:
For those that want to test, works way better on Firefox

Lercerss:
The new value seems a bit arbitrary, is there a specific reason you picked 10?

alvyn279:
> The new value seems a bit arbitrary, is there a specific reason you picked 10?

our xExtents property on the graph was blocking when the library was changing the array range on zoom out. giving it more ptsPerPxThreshold gave it a cushion to recalculate these ranges. The more the better, from what is seen from functional tests.

Issue #190, opened 2020-02-27
Service sometimes returns incorrect top of book


Lercerss:
Sometimes, the service returns incorrect points for the top-of-book graph, resulting in the best bid and ask intersecting. Clicking on the graph to investigate the faulty points shows the order book state is correct and the bids and asks do not cross, therefore this is an issue with the graph route and not the data. This only seems to occur in the small interval case when the service uses messages to yield points and not just orderbook snapshots.

Here is an example: 
![image](https://user-images.githubusercontent.com/34099471/75399499-5c18f900-58ca-11ea-8707-4010d16cc6d5.png)

Lercerss:
Seems to still occur for some time intervals with MSFT in prod. Further investigation is required.

PR #195, opened 2020-03-05
[#190] Fixed incorrect data in Top of Book graph

Lercerss:
- Removed assumption that orders are sorted by id within a price level
- Added test case to document


core/graphelier-service/models/orderbooks.go
```diff
@@ -174,7 +174,7 @@ func (orderbook *Orderbook) getOrderIndex(index int, message *Message) (int, boo
 
 	for ; i < len(orders); i++ {
 		lastID = orders[i].ID
-		if lastID >= message.OrderID {
+		if lastID == message.OrderID {
```
alvyn279: :astonished:

Issue #197, opened 2020-03-10
Setup Docker for the news API


chimano:
A Dockerfile needs to get setup for the news portion of the application.

## Subtasks
 - Write dockerfile for the news service (Application and database)
 - Modify CI to push and deploy the new docker images

chimano:
Merged in #198

PR #198, opened 2020-03-10
[#197] Set up docker for the news API

chimano:
- Set up basic docker files for the graphelier-news service
- Updated CI to include the new images

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,8 +11,8 @@ services:
     ports:
       - "5050:5050"
     depends_on:
-      - "mongo"
-  mongo:
+      - "service-db"
+  service-db:
     build:
       context: ./core/db
     container_name: "graphelier-db"
@@ -20,6 +20,22 @@ services:
       - "27017:27017"
     volumes:
       - "graphelier-db:/data/db"
+  graphelier-news:
+    build:
+      context: ./news/service
+    ports:
+      - "8080:8080"
+    depends_on:
+      - "news-db"
+  news-db:
+    build:
+      context: ./news/db
+    container_name: "news-db"
+    ports:
+      - "27016:27017"
+    volumes:
+      - "news-db:/data/db"
 
 volumes:
   graphelier-db:
+  news-db:
--- a/news/.gitignore
+++ b/news/.gitignore
@@ -1,0 +1,3 @@
+venv
+.vscode
+__pycache__
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,10 +66,10 @@ jobs:
             - run:
                   name: Test service
                   command: |
-                    mkdir -p /tmp/artifacts
-                    go test --coverprofile="go-cover.out" -v graphelier/core/graphelier-service/...
-                    go tool cover -html="go-cover.out" -o coverage.html
-                    mv coverage.html /tmp/artifacts
+                      mkdir -p /tmp/artifacts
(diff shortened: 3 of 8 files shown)


core/graphelier_news/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM python:3.7.4 as builder
+WORKDIR /graphelier_news
+COPY requirements.txt .
+RUN pip install -r requirements.txt
+COPY . .
+
+FROM builder as lint
+CMD [ "pylint", "scripts" ]
```
Lercerss: Should probably be 
```suggestion
CMD [ "pylint", "graphelier_news" ]
```

core/graphelier_news/requirements.txt
```diff
@@ -0,0 +1,10 @@
+astroid==2.3.3
+autopep8==1.5
+isort==4.3.21
+lazy-object-proxy==1.4.3
+mccabe==0.6.1
+nltk==3.4.5
+pycodestyle==2.5.0
+pylint==2.4.4
+six==1.14.0
+wrapt==1.11.2
```
alvyn279: flask ?

docker-compose.yml
```diff
@@ -11,15 +11,27 @@ services:
...
+    ports:
+      - "8080:8080"
+    depends_on:
+      - "news-db"
+  news-db:
+    build:
+      context: ./core/news-db
+    container_name: "news-db"
+    ports:
+      - "27016:27017"
```
alvyn279: port 27017 of the mongo news-db container to public 27016 right ?

chimano: yep

core/graphelier_news/requirements.txt
chimano: well I didn't use flask for the setup so i didn't see the need to install it now

docker-compose.yml
```diff
@@ -11,15 +11,27 @@ services:
     ports:
       - "5050:5050"
     depends_on:
-      - "mongo"
-  mongo:
+      - "graphelier-db"
```
Lercerss: How does this work with volumes?

chimano: > Can we rename `core/graphelier_news` to `core/graphelier-news` to match `core/graphelier-service`?
Well since graphelier_news is a python module i wasn't sure if naming it graphelier-news would work

Lercerss: Since this module won't be imported directly I don't think that should be a problem, I don't have a strong preference on the exact name I would just like to have the naming conventions be consistent with existing directories.

chimano: it will be for linting.

`CMD [ "pylint", "graphelier_news" ]`

Lercerss: Tested it out locally with a package called `test-dash`, added an `__init__.py` and ran `pylint test-dash` without issues.

Lercerss: The only problem would arise if we need to import something from outside the `graphelier-news` directory, say from the `scripts`. Then the solution would be to use string imports with `graphelier_news = __import__("graphelier-news")`

chimano: As for the volume, it would use a new one

news/.gitignore
```diff
@@ -0,0 +1,3 @@
+venv
+.vscode
+__pycache__
```
Lercerss:
```suggestion
**/__pycache__
```
Pycache directories will be created in all subdirectories when running the module

chimano: Subdirectories are already covered the way it is

Issue #188, opened 2020-02-20
Click on messages in order information drawer


Lercerss:
Add navigation possibilities for users to jump to messages in an order's lifecycle.

Story #110
## Tasks
- [x] Clicking on a message inside the drawer should go to that message's time

riyaGit:
Addressed in pull request #205

PR #205, opened 2020-04-01
#188 Clicking on message in Order Details

riyaGit:
Screen Recording 2020-03-31 at 11 06 42 PM


app/src/components/OrderInformation.tsx
```diff
@@ -102,6 +125,8 @@ class OrderInformation extends Component<Props, State> {
                             <TableRow
                                 key={sod_offset}
                                 id={'orderMessageListRow'}
+                                className={classes.orderMessageListRow}
+                                onClick={() => this.handleOnMessageClick(timestamp)}
```
alvyn279, line 129: can there be an indication of which table row we're hovering over?

Issue #208, opened 2020-04-02
Group related news articles


chimano:
User story #163 

Cluster related articles together.

Articles that are similar to one another should be grouped together to avoid relaying duplicate information

chimano:
implemented in #212

PR #212, opened 2020-04-03
[#208] Clustering of news articles

chimano:
- News articles are now clustered using TFIDF
and agglomerative clustering


news/scripts/news_cluster.spec.py
```diff
@@ -0,0 +1,124 @@
...
+        "timestamp" : "1583554020"
+    }
+    # END OF CLUSTER 2
+]
+with description('News Cluster:') as self:
+    with it('should create 2 clusters'):
+        with patch('news_cluster.fetch_all') as mock_fetch, patch('news_cluster.insert_cluster') as mock_insert:
```
alvyn279, line 98: nice didn't know you could do this

news/scripts/news_cluster.py
```diff
@@ -0,0 +1,93 @@
...
+    parser = script_args()
+    args = parser.parse_args()
+    pipeline = ArticleClustering(args.threshold)
+    pipeline.run()
+
+
+def script_args():
+    parser = argparse.ArgumentParser(
+        description='Articles fetcher for StockNewsAPI', formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument('--threshold', help='The maximum distance between articles that will be accepted in the cluster. Increasing this number decreases the size of clusters (Default=1.4)',
```
Lercerss, line 27: The `ArgumentDefaultsHelpFormatter` will add that in for you.
```suggestion
    parser.add_argument('--threshold', help='The maximum distance between articles that will be accepted in the cluster. Increasing this number decreases the size of clusters',
```

news/scripts/news_cluster.py
```diff
@@ -0,0 +1,93 @@
...
+
+from models.news import Cluster
+from mongo_news_db.db_connector import (delete_clusters, fetch_all,
+                                        insert_cluster)
+
+SECONDS_IN_DAY = 24 * 3600
+
+
+def main():
+    delete_clusters()
```
Lercerss, line 16: This should be done after parsing the arguments. Right now calling `python news_cluster.py -h` will wipe your db

news/scripts/news_cluster.spec.py
```diff
@@ -0,0 +1,124 @@
+from unittest.mock import patch
+
+from mamba import description, it
+from sure import expect  # pylint: disable=W0611
+
+from news_cluster import ArticleClustering
+
+DATA = [ {
```
Lercerss, line 8: Can we move these constants into a fixtures file?

chimano: I don't see the utility behind it if the data will only be used in this test suite

news/scripts/news_cluster.py
chimano: good catch

news/scripts/news_cluster.spec.py
Lercerss: I just find it obscures the test behavior

chimano: how?

Lercerss: The majority of the file's content is the data, while the tests are lost at the bottom. I thought it would be clearer to separate the data from the expected behavior

chimano: Well it's a .spec.py file, so it's easily identifiable as a test file. I don't think the behavior gets obscured just by the presence of data.

Issue #178, opened 2020-02-05
Serve modifications to the order book for playback


Lercerss:
From the service perspective, we need to support a low-latency interface to serve modifications to the order book over time. This could be done by first generating an order book snapshot, then using the following messages to extract modifications over time. Message types should be translated into more generic modification types and the necessary information to animate the modification should be extracted.

We should investigate Lobster's data to see if we can support modification of an order's price by detecting a delete followed by a new, for the same quantity at exactly the same timestamp.

# Tasks
- Determine low-latency solution, possibly using websockets and grouping modifications by time intervals
- Implement algorithm to translate messages to order book modifications ready for animation
- Serve modifications using a synchronous or asynchronous strategy

Story #175.

PR #214, opened 2020-04-08
[#178] Added last_sod_offset to playback response

Lercerss:
- Added last_sod_offset to playback response
- Fixed rateRealtime usage, mult. instead of div. of delay

### DIFF
--- a/core/graphelier-service/api/hndlrs/playbackhandler.go
+++ b/core/graphelier-service/api/hndlrs/playbackhandler.go
@@ -80,7 +80,7 @@ func StreamPlayback(env *Env, w http.ResponseWriter, r *http.Request) error {
 	case mErr == nil && tErr != nil:
 		loader = &CountIntervalLoader{Instrument: instrument, Datastore: env.Datastore, Count: rateMessages}
 	case tErr == nil && mErr != nil:
-		loader = &TimeIntervalLoader{Instrument: instrument, Datastore: env.Datastore, Interval: uint64(delay / rateRealtime)}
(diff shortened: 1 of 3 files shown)


core/graphelier-service/api/hndlrs/playbackhandler.go
```diff
@@ -80,7 +80,7 @@ func StreamPlayback(env *Env, w http.ResponseWriter, r *http.Request) error {
 	case mErr == nil && tErr != nil:
 		loader = &CountIntervalLoader{Instrument: instrument, Datastore: env.Datastore, Count: rateMessages}
 	case tErr == nil && mErr != nil:
-		loader = &TimeIntervalLoader{Instrument: instrument, Datastore: env.Datastore, Interval: uint64(delay / rateRealtime)}
+		loader = &TimeIntervalLoader{Instrument: instrument, Datastore: env.Datastore, Interval: uint64(delay * rateRealtime)}
```
alvyn279, line 83: seems to be a pretty big change to the logic, what's happening here?

Lercerss: Completely agree, rateRealtime is supposed to be a ratio corresponding to how much faster/slower we should be playing back compared to real time. Say the request specifies `rateRealtime=0.5`, then we should be playing back at 50% real time speed. To determine the interval at which to load more messages, we should multiply with the delay instead of dividing. For example with a delay of 4s and a rateRealtime of 0.75, we should be sending 3 seconds of data every 4 seconds.

PR #219, opened 2020-04-09
Kontrolol/playback

chris-baur:
FE playback features.


app/src/components/OrderBookSnapshot.tsx
```diff
@@ -256,6 +265,110 @@ class OrderBookSnapshot extends Component<Props, State> {
...
+        const modificationsLength = playbackData.modifications.length;
+        let ctr: number = 0;
+        const data = processOrderBookPlayback(listItems);
+        let { newListItems } = data;
+        const { newMaxQuantity } = data;
+        while (ctr < modificationsLength) {
+            const playbackModification = playbackData.modifications[ctr];
+            const { price } = playbackModification;
+            const { from } = playbackModification;
+            const { to } = playbackModification;
```
alvyn279, line 288: extraction can be done in a single line ? `{price, from, to} = playbackModification`

app/src/components/TimestampOrderBookScroller.tsx
```diff
@@ -68,36 +70,48 @@ class TimestampOrderBookScroller extends Component<Props> {
...
-        });
+        const { playback } = this.props;
+        if (!playback) {
+            this.middleReferenceItem
+            && this.middleReferenceItem.current
+            && this.middleReferenceItem.current.scrollIntoView({
+                behavior: 'smooth',
+                block: 'center',
+            });
+        }
```
alvyn279, line 92: I think the button should be visually disabled if this feature is not accessible

app/src/components/OrderBookSnapshot.tsx
chris-baur: 👍

app/src/components/TimestampOrderBookScroller.tsx
chris-baur: ok, will make it disabled.

chris-baur:
> 
> 
> This my first quick pass over this pr, since it's really big. functionality is looking good
> 
> I also think viewing the order details on a specific order should be disabled during playback.

Addressed in local changes. Also disabled message navigation buttons instead of if in handler method.

app/src/components/PlaybackControl.tsx
```diff
@@ -0,0 +1,266 @@
...
+    /**
+     * @desc Calls the backend service to get new data to feed OrderBook, graph
+     */
+    getPlaybackOrderBookData = (): void => {
+        const {
+            selectedInstrument, lastSodOffset, handlePlaybackModifications,
+        } = this.props;
+        const parameter = this.getPlaybackParameter();
+
+        const endPoint = `ws://localhost:5050/playback/${selectedInstrument}/${lastSodOffset}/${parameter}`;
```
chimano, line 158: the endpoint should be in the constants.js file so that we can make the app point to a host by only changing a single line

app/src/components/PlaybackControl.tsx
```diff
@@ -0,0 +1,266 @@
...
+            console.log('opened playback websocket');
+        };
+        this.playbackWS.onmessage = m => {
+            const data: PlaybackData = JSON.parse(m.data);
+            handlePlaybackModifications(data);
+        };
+        this.playbackWS.onclose = () => {
+            console.log('closed playback websocket');
+        };
+    };
```
chimano, line 171: Should we put this code in some service file?

app/src/components/PlaybackControl.tsx
Lercerss: Please follow the existing pattern for backend communication from the frontend by adding a function in the service class

Lercerss:
Can we reposition the playback buttons to be at the same height as the instrument/date selection?
The whitespace above them feels off.
image

app/src/components/PlaybackControl.tsx
```diff
@@ -0,0 +1,266 @@
...
+                            );
+                        })}
+                    </Select>
+                </FormControl>
+                <TextField
+                    id={'unitSpeedSelector'}
+                    type={'number'}
+                    value={unitSpeed}
+                    onChange={this.handleUnitSpeedChange}
+                    label={'# of Units'}
```
Lercerss, line 255: Can you make this a dynamic label based on the unit selected? I don't feel this is descriptive enough

app/src/components/PlaybackControl.tsx
```diff
@@ -0,0 +1,266 @@
...
+        }
+        return (timeInNano / NANOSECONDS_IN_ONE_SECOND) * unitSpeed;
+    };
+
+    /**
+     * @desc Gets playback speed parameter based on selected unit and unit speed
+     */
+    getPlaybackParameter = (): string => {
+        const { unitSpeed, selectedUnit } = this.state;
+        let parameter: string = '?delay=2.5&';
```
Lercerss, line 141: 2.5s delay is too noticeable from a user's perspective. I tried 0.5s locally and it seems the app can handle it well. To account for this, when the user selects "Messsages" as a unit, let's multiply the input number by the delay so that they always see "X Messages per second".

app/src/components/PlaybackControl.tsx
chris-baur: Im not too sure, cause normal calls just make a request through the service and then handle the data with some processing and set states

Lercerss: I still think the handlers should be in the service file. Especially since you are not using onclose/onopen, then you would only need to pass in the onmessage handler as a parameter

chris-baur: But then what do you suggest for closing the web socket ? we put it as a static variable ?

chris-baur: Currently there is axios logging for HTTP requests, do we want to remove the console logs for opening and closing the socket?

chimano: Perhaps you can store it as a class variable like this guy: https://dev.to/finallynero/using-websockets-in-react-4fkp

chris-baur: thsts what i did initially. it was a class var in playback

chris-baur: now its staitc in orderbook service

Lercerss:
It seems to be due to the message list loading more messages, though I cannot tell for sure

alvyn279:
> It seems to be due to the message list loading more messages, though I cannot tell for sure

I don't get that issue during my testing, the only thing that seems to be off is the selected message not being scrolled to.  @chris-baur, maybe fixing the latter will improve things overall ?

chris-baur:
> > It seems to be due to the message list loading more messages, though I cannot tell for sure
> 
> I don't get that issue during my testing, the only thing that seems to be off is the selected message not being scrolled to at each step. @chris-baur, maybe fixing the latter will improve things overall ?

I Purposely did not let messages scroll to current message during playback. I can re enable it if thats desired

chris-baur:
I left centering to currently applied message in the message list off when playback is on. You can still manually center if you want. Reason is, is that on smalled screens (anyhtign not 4) you would even have trouble trying to pause the playback since the centering method used for top of book or message list makes your screen also scroll back to the element which has been centered.
Example of what we do not want
[Uploading 2020-04-13 00-26-12.zip…]()

alvyn279:
> I left centering to currently applied message in the message list off when playback is on. You can still manually center if you want. Reason is, is that on smalled screens (anyhtign not 4) you would even have trouble trying to pause the playback since the centering method used for top of book or message list makes your screen also scroll back to the element which has been centered.
> Example of what we do not want
> Uploading 2020-04-13 00-26-12.zip…

cool forget about the centering i proposed then

Issue #201, opened 2020-03-12
Serve processed news data


Lercerss:
We need a new service to provide processed news data for the timeline. It should be separate from our existing Go backend to avoid tangling concerns. To reduce development time, we could build it in Python using Flask.

Related story: #163 

## Tasks
- Initialize Flask server code base in news directory
- Connect news service to the news mongo db
- Add endpoints to read, process and serve data for the frontend timeline

alvyn279:
Implemented in #220 and #222

alvyn279:
Tests for new news clusters route:
https://github.com/Lercerss/graphelier/blob/master/news/service/views.spec.py

PR #220, opened 2020-04-09
[#201] Implemented route for news cluster feature

alvyn279:
- Modified sample flask application to support a root route and a news clusters route
- Added tests
- Sample news cluster route usage:
`http://localhost:8080/news_clusters/1581735096?direction=1&quantity=20`
- Sample response:
```
{
  "clusters": [{
      "articles": [
        {
          "article_url": "https://www.fool.com/investing/2020/02/18/better-buy-apple-vs-google.aspx", 
          "date": "Tue, 18 Feb 2020 07:45:00 -0500", 
          "image_url": "https://cdn.snapi.dev/images/v1/k/8/stocks1-2.jpg", 
          "sentiment": "Neutral", 
          "source_name": "The Motley Fool", 
          "summary": "Size isn't necessarily everything to investors, of course. The case for, and against, Apple It's a name that needs no introduction. But there's the rub, so to speak. In the meantime, YouTube's advertising prowess is tremendous. It generated $4.7 billion in revenue last quarter just by showing ads before, during, or after videos. It's not an easy choice. Tax implications must always be considered, and if any stock can defy the odds, it's Apple.", 
          "tickers": [
            "AAPL", 
            "GOOG", 
            "GOOGL"
          ], 
          "timestamp": "1582029900", 
          "title": "Better Buy: Apple vs. Google"
        }
      ], 
      "size": 1, 
      "timestamp": 1582029900
    }, 
    {
      "articles": [
        {
          "article_url": "https://www.marketwatch.com/story/apple-set-to-lose-title-of-most-valuable-us-companies-to-microsoft-2020-02-18", 
          "date": "Tue, 18 Feb 2020 07:57:03 -0500", 
          "image_url": "https://cdn.snapi.dev/images/v1/1/0/103922274-rtx2ojun-3.jpg", 
          "sentiment": "Negative", 
          "source_name": "Market Watch", 
          "summary": "With Apple Inc. shares AAPL, +0.27% falling 3.1% in premarket trading Tuesday, Microsoft Corp. MSFT, -0.15% is on track to take back the No. 1 position on the list of the U.S.'s most valuable companies by market capitalization. Microsoft's stock is down 0.2% ahead of the open, which means the software giant is headed toward a market cap of $1.408 trillion at the open, while Apple is on track to open with a market cap of $1.378 trillion. The last time Microsoft closed as the most valuable company was Feb. 11 and Feb. 10, which ended a 67-session run with Apple at No. 1. The companies were still well ahead of third-place Amazon.com Inc. AMZN, +0.41% , which closed Friday with at market cap of $1.063 trillion, and fourth-place Alphabet Inc. GOOGL, +1.46% at $1.044 trillion.", 
          "tickers": [
            "AAPL"
          ], 
          "timestamp": "1582030623", 
          "title": "Apple set to lose title of most valuable U.S. companies to Microsoft"
        }
      ], 
      "size": 1, 
      "timestamp": 1582030623
    }
  ], 
  "next_timestamp": 1582031280, 
  "timestamp": 1581735096
}
```


news/service/db.py
```diff
@@ -0,0 +1,18 @@
+from pymongo import MongoClient
+
+_DB_CLIENT = MongoClient("mongodb://news-db:27017/news-db")
+_DB_CLUSTERED_ARTICLES = _DB_CLIENT['news-db']['clusteredarticles']
+
+
+def fetch_n_clusters_before_time(timestamp: int, n_clusters: int):
+    return _DB_CLUSTERED_ARTICLES \
+        .find({'timestamp': {'$lt': timestamp}}) \
+        .sort([('timestamp', -1)]) \
```
chimano, line 10: why does sort direction change?

news/service/db.py
```diff
@@ -0,0 +1,18 @@
+from pymongo import MongoClient
+
+_DB_CLIENT = MongoClient("mongodb://news-db:27017/news-db")
+_DB_CLUSTERED_ARTICLES = _DB_CLIENT['news-db']['clusteredarticles']
+
+
+def fetch_n_clusters_before_time(timestamp: int, n_clusters: int):
```
chimano, line 7: i feel like both methods can be combined into one

news/service/views.py
```diff
@@ -0,0 +1,37 @@
...
+def news_clusters(start_timestamp):
+    direction = int(request.args.get('direction', '1'))
+    quantity = int(request.args.get('quantity', '10'))
+    try:
+        timestamp = int(start_timestamp)
+    except ValueError as e:
+        LOGGER.error(e)
+        return jsonify({'message': 'Given timestamp is badly formed: {}'.format(start_timestamp)}), 400
+
+    query_result = fetch_n_clusters_after_time(timestamp, quantity+1) if direction == 1 \
```
chimano, line 23: why +1 on the quantity?

news/service/views.py
```diff
@@ -0,0 +1,37 @@
...
+        else fetch_n_clusters_before_time(timestamp, quantity+1)
+    clusters = [cluster_cursor_to_dict(cluster_cursor) for cluster_cursor in query_result]
+    LOGGER.info('Found {} clusters'.format(len(clusters)))
+    next_timestamp = timestamp
+
+    if len(clusters) != 0:
+        next_timestamp = clusters[len(clusters) - 1]['timestamp']
```
chimano, line 30: clusters[-1] gets you the last element too

news/service/views.py
alvyn279: I want to know the timestamp of the (last element + 1) of the query such that the user can know which timestamp to request next in terms of paging (`next_timestamp`).

Basically the user holds and updates the upper `next_timestamp` and `lower next_timestamp` for paging on client side

news/service/db.py
alvyn279: sure, i just wanted to make it more verbose

news/service/db.py
alvyn279: to be able to locate `next_timestamp` in resulting query, as explained above

news/service/views.py
chimano: but they would be able to do that with the first and last element they have on their end no?

alvyn279: then they'll receive duplicated values for those elements when requesting with their respective timestamps

alvyn279: oh i'm doing a `lt` / `gt`, shouldn't happen right ?

chimano: that shouldn't be the case because the queries are lt/gt and not lte/gte

news/service/db.py
alvyn279: actually my last comment is wrong, I did it because I want to find the X clusters happening before timestamp. `.limit` is going to take the first X encountered in the query result, hence the reversed sorting

chimano: Ah that makes sense. You should sort the results in python then so that timestamp is always increasing

alvyn279:
> Looks good, whatever I had noticed Chirac brought up already.
> Just wondering if there were constants used in BE for making it more verbose instead of -1 and 1

I think 1 going forward in time and -1 going backwards in time speaks for itself. But good point, im actually gonna include a docstring

docker-compose.override.yml
```diff
@@ -12,3 +12,8 @@ services:
       target: dev
     volumes:
       - ./core/graphelier-service:/go/src/graphelier/core/graphelier-service
+  graphelier-news:
+    build:
+      target: dev
+    volumes:
+      - ./news/service:/news/service
```
Lercerss, line 19: Since there's no hot-reload setup for dev I dont see why this is done

news/service/Dockerfile
```diff
@@ -1,10 +1,11 @@
...
+RUN pip install -r requirements.txt
 
 FROM builder AS dev
+ENV FLASK_APP=server.py
+ENV FLASK_DEBUG=1
+ENV FLASK_ENV=development
 EXPOSE 8080
-ENTRYPOINT [ "python" ]
-CMD ["main.py"]
+CMD ["python", "server.py"]
```
Lercerss, line 11: Could you add a prod entrypoint? The only way to run the news-service with your current setup is to use the dev docker-compose file

docker-compose.override.yml
alvyn279: i set up the hot reload

Lercerss:
Could you unify the types for `timestamp` in the JSON response? You have both string and number

PR #222, opened 2020-04-10
[#201] Fix for exposed port on prod for graphelier news

alvyn279:
Fix for a port issue on prod.

### DIFF
--- a/news/service/Dockerfile
+++ b/news/service/Dockerfile
@@ -8,7 +8,7 @@ ENV FLASK_APP=server.py
 ENV FLASK_DEBUG=1
 ENV FLASK_ENV=development
 EXPOSE 8080
-CMD ["python", "server.py"]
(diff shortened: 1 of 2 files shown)


news/service/Dockerfile
```diff
@@ -8,7 +8,7 @@ ENV FLASK_APP=server.py
 ENV FLASK_DEBUG=1
 ENV FLASK_ENV=development
 EXPOSE 8080
-CMD ["python", "server.py"]
+CMD [ "python", "-m", "flask", "run", "--host", "0.0.0.0", "--port", "8080" ]
```
Lercerss, line 11: How will this affect dev/hot-reload?

alvyn279: still works fine :)

Lercerss: Interesting, is `server.py` still used?

alvyn279: yeah through env variable `FLASK_APP=server.py`

Issue #193, opened 2020-02-28
Add News Timeline Page


panoreak:
Add a page to the app that allows users to view news articles on a timeline:

![timeline](https://user-images.githubusercontent.com/25068506/75583985-b8038f00-5a3c-11ea-856a-b56b3e44bcf0.png)

# Tasks
- Find scrollable timeline library
- Add News Timeline page to the app
- Create component for the news article items. Component should allow user to choose to view the full article, or view the TOB graph of an instrument linked to the article, with the selected time set to the article's published time
- When the page is loaded, it must populate the timeline with the most recent news articles available from the backend route
- When the timeline is scrolled, the page should request additional news articles (must disable scrolling into the future)

Story: #163

PR #224, opened 2020-04-12
[#193] News Timeline Frontend

panoreak:
Added news timeline page that allows the user to browse stock-related news articles in January and February 2020.

graphelier-Google-Chrome-2020-04

### DIFF
--- a/app/src/components/ArticleDetails.tsx
+++ b/app/src/components/ArticleDetails.tsx
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import { withStyles } from '@material-ui/core/styles';
 import { createStyles, WithStyles } from '@material-ui/styles';
-import classNames from 'classnames';
 
 import Img from 'react-image';
 import {
@@ -11,15 +10,18 @@ import {
 import OpenInNewIcon from '@material-ui/icons/OpenInNew';
 import { Link } from 'react-router-dom';
 import bigInt from 'big-integer';
+import { mdiEmoticonHappyOutline, mdiEmoticonNeutralOutline, mdiEmoticonSadOutline } from '@mdi/js';
+import Icon from '@mdi/react';
 import { Styles } from '../styles/ArticleDetails';
 import {
     Article,
 } from '../models/NewsTimeline';
-import { NANOSECONDS_IN_ONE_SECOND, ORDERBOOK_INSTRUMENTS } from '../constants/Constants';
+import { INSTR_COLOR, NANOSECONDS_IN_ONE_SECOND } from '../constants/Constants';
 import {
     getHoursMinutesStringFromTimestamp,
     getDateStringFromTimestamp,
 } from '../utils/date-utils';
+import { Colors } from '../styles/App';
 
 const styles = createStyles(Styles);
 
@@ -38,8 +40,23 @@ class ArticleDetails extends Component<Props, State> {
         const datePublishedString = getDateStringFromTimestamp(nanosecondTimestamp);
 
         const instruments = article.tickers.filter(ticker => {
-            return ORDERBOOK_INSTRUMENTS.includes(ticker);
+            return ticker in INSTR_COLOR;
         });
+
+        const sentimentOptions = {
+            Positive: {
+                path: mdiEmoticonHappyOutline,
+                color: Colors.positiveGreen,
+            },
+            Neutral: {
+                path: mdiEmoticonNeutralOutline,
+                color: Colors.neutralGrey,
+            },
+            Negative: {
+                path: mdiEmoticonSadOutline,
+                color: Colors.negativeRed,
+            },
+        };
 
         return (
             <div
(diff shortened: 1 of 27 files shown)


app/src/constants/Constants.ts
```diff
@@ -1,13 +1,16 @@
 import bigInt from 'big-integer';
+import { Colors } from '../styles/App';
 
 export const ENVIRONMENT = 'DEV';
 export const APP_NAME = 'Graphelier';
-export const BACKEND_PORT = 5050;
-let host = 'localhost';
+export const ORDERBOOK_SERVICE_BACKEND_PORT = 5050;
+export const NEWS_SERVICE_BACKEND_PORT = 8080;
+let host = '18.218.121.174';
```
Lercerss, line 8: You let this slip :)

panoreak: Yikes! You're right. Fixing it sap.

panoreak:
> I get this issue while testing out locally, for a large cluster it seems?
> image
> Also, is it expected that the timeline doesnt seem to occupy the maximum amount of screen space? The left, right and bottom borders are not quite at the edge of the screen making me wonder how this would look like on a 4K display



> I get this issue while testing out locally, for a large cluster it seems?
> image
> Also, is it expected that the timeline doesnt seem to occupy the maximum amount of screen space? The left, right and bottom borders are not quite at the edge of the screen making me wonder how this would look like on a 4K display

There's not much i can do about the width. Even setting the width to 100% does not change the layout. Also, since the <main /> container is not a fixed height, I must set the height of the MultiDirectionalScroller containing the timeline to a fixed height, and I can't use 100% height. Therefore, I have the height set to 85vh. Modifying the <main /> container to be a fixed height, would have side effects on OrderBookSnapshot, which is also in <main />.

chris-baur:
> > I get this issue while testing out locally, for a large cluster it seems?
> > image
> > Also, is it expected that the timeline doesnt seem to occupy the maximum amount of screen space? The left, right and bottom borders are not quite at the edge of the screen making me wonder how this would look like on a 4K display
> 
> > I get this issue while testing out locally, for a large cluster it seems?
> > image
> > Also, is it expected that the timeline doesnt seem to occupy the maximum amount of screen space? The left, right and bottom borders are not quite at the edge of the screen making me wonder how this would look like on a 4K display
> 
> There's not much I can do about the width. Even setting the width to 100% does not change the layout. Also, since the `<main />` container is not a fixed height, I must set the height of the MultiDirectionalScroller containing the timeline to a fixed height, and I can't use 100% height. Therefore, I have the height set to 85vh. Modifying the `<main />` container to be a fixed heightwould have side effects on OrderBookSnapshot, which is also in `<main />`.

setting width to 100% is gonna put it to a width of the current screen it is on. It needs to be within some kind of container that allows the overdlow of width ( x ) to be scrollable

app/src/constants/Constants.ts
```diff
@@ -29,3 +32,18 @@ export const NANOSECONDS_IN_ONE_DAY = NANOSECONDS_IN_ONE_HOUR * 24;
...
+    MSFT: {
+        backgroundColor: Colors.emerald,
+    },
+    SPY: {
+        backgroundColor: Colors.orange,
+    },
+    default: {
+        backgroundColor: Colors.grey,
+    },
+};
```
chimano, line 49: Instruments shouldn't be hard coded, they should be fetched on the backend

app/src/components/ArticleItem.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    render() {
+        const { article, classes, isFirstItem } = this.props;
+        const { modalIsOpen } = this.state;
+        const nanosecondTimestamp = bigInt(article.timestamp).multiply(NANOSECONDS_IN_ONE_SECOND);
+        const timePublishedString = getHoursMinutesStringFromTimestamp(nanosecondTimestamp);
+
+        const instruments = article.tickers.filter(ticker => {
+            return ticker in INSTR_COLOR;
+        });
```
chimano, line 65: We want all tickers to appear not just the ones we have on the backend

app/src/constants/Constants.ts
panoreak: If we want the tickers to always have the same color, how else can we proceed?

app/src/components/ArticleItem.tsx
panoreak: Sorry about that, I thought we didn't want to include them. For most articles, the number of tickers was pretty big, so I thought it looked cluttered. I'll start working on this.

chimano: How big does it get?

alvyn279: i think you can get up to 6-7 at times, making the line cluttered with instrument pills.

app/src/constants/Constants.ts
alvyn279: i think saving the instruments to redux would be a good way to serve all our use cases.

alvyn279: the fetched instruments, that is

app/src/components/ArticleItem.tsx
panoreak: I just pushed a commit @chimano, you can see all the tickers now if you want to see for yourself

Lercerss: Looks good!

alvyn279:
thanks for the css change @chimano, works beautifully

app/src/components/ArticleDetails.tsx
```diff
@@ -0,0 +1,156 @@
...
+class ArticleDetails extends Component<Props, State> {
+    render() {
+        const { article, classes, instruments } = this.props;
+
+        const nanosecondTimestamp = bigInt(article.timestamp).multiply(NANOSECONDS_IN_ONE_SECOND);
+        const timePublishedString = getHoursMinutesStringFromTimestamp(nanosecondTimestamp);
+        const datePublishedString = getDateStringFromTimestamp(nanosecondTimestamp);
+        const orderedInstruments = article.tickers.sort(a => {
+            return instruments.includes(a) ? 0 : 1;
+        });
```
alvyn279, line 46: @chimano since you sort on the article's tickers already at parent level, this sort isn't necessary i think

chimano: Ah i hadn't noticed that repeat

Issue #210, opened 2020-04-02
Capstone website video


Lercerss:
As a developer, I want to produce a video to upload on the capstone website.

| Metric       | Value           |
|--------------|-----------------|
| Story Points | 5 |
| Priority     | High |
| Risk         | Low |

# Tasks
- Plan features to demo and content of video
- Record footage
- Edit
- Upload

PR #227, opened 2020-04-13
[#210] Graphelier Logo 

riyaGit:
The following logo was created and added to the app
graphelierWhiteLogo4

Screen Shot 2020-04-13 at 1 04 13 AM


riyaGit:
> The picture should be in app/src/assets/

Done!

